PR #71, opened 2014-12-20
Adds support for 64-bit integer and some float operations for 32-bit x86...

acw:
....

On i386 platforms, modern compilers are emitting call-outs to 64-bit integer
arithmetic. This patch adds support for these functions. Implementations are
taken from LLVM's compiler-rt library, which appears to have a compatible
license with openlibm. Appropriate license information has been added to
LICENSE.md to reflect this merge.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -13,6 +13,13 @@ OpenLibm library from msun, as part of the Julia project is covered
 under the MIT license. The test files, test-double.c and test-float.c
 are under the LGPL.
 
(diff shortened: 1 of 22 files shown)


acw:
(Sorry, missed your original comments.)

Actually, the idea was just as described: call-outs to these functions are generated by modern compilers, and I thought libm would be a good place to collect implementations for projects that don't use libgcc and etc. (Mirage, HaLVM, etc.)

I can shift them over somewhere else within my project, but openlibm seemed like a more ideal home.

PR #72, opened 2015-01-05
Fix build on ARM and add Travis tests

talex5:
Recent changes had broken the build on ARM by adding more long double functions. These should be skipped on ARM, where double = long double.

This commit also adds Travis CI tests, which can be run automatically on every push to github. To do that, go to https://travis-ci.org/ and enable tests for the repository (it's free). Ideally, do this before merging (it will only run the tests for commits made after enabling it).

Here's the output from my fork (which has tests enabled): https://travis-ci.org/talex5/openlibm/builds/45933733

It does two builds on each commit, one on x86 and one cross-compiling to arm (which would detect the bug fixed above). On x86, it also runs the unit-tests. As the unit-tests currently fail on ARM I have disabled them, but if you set LOADER to `ld-linux-armhf.so.3` in the `.travis.sh` file (it's commented out at the moment) then it will run them too, if you want to debug it.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+set -eux
+
+case "$TARGET" in
+host)
+	uname -a
+	export LOADER=
+	make ;;
+arm32)
+	sudo bash -c 'echo >> /etc/apt/sources.list "deb http://archive.ubuntu.com/ubuntu/ trusty main restricted universe"'
+	sudo apt-get update
+	sudo apt-get -y install gcc-4.7-arm-linux-gnueabihf qemu binfmt-support
+	make CC="arm-linux-gnueabihf-gcc-4.7"
+	export LD_LIBRARY_PATH=/usr/arm-linux-gnueabihf/lib
+	#export LOADER=/usr/arm-linux-gnueabihf/lib/ld-linux-armhf.so.3
+	export LOADER="echo TESTS DISABLED ON ARM"
+	;;
+*)
+	echo 'Unknown TARGET!'
+	exit 1
+	;;
+esac
+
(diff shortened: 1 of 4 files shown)


ViralBShah:
Thanks - I had been meaning to get around to testing this for a while.

ViralBShah:
We really do need much better unit testing for openlibm. I am not a fan of the current tests, but that is better than having nothing.

ViralBShah:
How will it run the ARM tests on an x86 box?

ViralBShah:
BTW, any idea how to enable the tests for osx? I think I don't fully understand the travis configuration.

talex5:
It uses qemu to run the ARM tests on x86. I think OS X support on Travis is on some kind of limited beta, and not generally available.

tkelman:
@ViralBShah see the Julia repo, you need an `os:` key. And email support@travis-ci.com to ask for them to enable it on this repository. Various packages have been enabling it recently, apparently the turnaround's not too bad right now http://docs.travis-ci.com/user/multi-os/

tkelman:
Oh you already did in 769a422f06faa0b859f64dddd6165f5ddb32efb7, nevermind. You probably want to exclude the arm build for osx, similar to how I did here https://github.com/tkelman/julia/commit/a38da6ca03180ed0b0111409a2e804b6555e3f54

ViralBShah:
Will this do the trick:

````
language: c
script: ./.travis.sh
os:
    - linux
    - osx
env:
    - TARGET=host
    - TARGET=arm32
matrix:
    exclude:
        - os: osx
        env: TARGET="arm32"
notifications:
    email: false

````

tkelman:
I believe so

PR #73, opened 2015-01-07
add -f flag to cp in install target

tkelman:
Would anyone be opposed to this? I get occasional build failures in Julia due to "unable to cp, file exists" or similar errors. This was probably triggered by a different change I made in one of Julia's makefiles, but that was fixing a different problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ env:
 matrix:
     exclude:
         - os: osx
-          env: TARGET="arm32"
+          env: TARGET=arm32
 notifications:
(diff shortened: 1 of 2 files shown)


tkelman:
Looks like the osx tests got turned on, and they fail. cc @viralbshah

ViralBShah:
I wrote to them as you suggested and they did it. The tests have been in that condition for a while - a few ULPs here and there. Could be a gcc/clang difference or something else. At least enough digging has to now be done to make the tests pass.

PR #74, opened 2015-01-07
Portability fixes

EdSchouten:
Hi there,

I'm currently working on getting openlibm working in an alternative POSIX-like environment which I'm planning on releasing in a couple of months from now. It seems the code builds pretty well, though I needed to tweak it in a couple of places.

Attached are a couple of cleanups that are needed to get openlibm to build.

Ed

### DIFF
--- a/amd64/bsd_ieeefp.h
+++ b/amd64/bsd_ieeefp.h
@@ -38,10 +38,6 @@
 #ifndef _MACHINE_IEEEFP_H_
 #define _MACHINE_IEEEFP_H_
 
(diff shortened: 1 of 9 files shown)


EdSchouten:
It seems that even though the code builds on OS X, the tests fail. My suspicion is that this is unrelated to this change, as the previous build also failed this way.

Keno:
Yes, the OS X tests are borked.

ViralBShah:
Thanks. These are a nice set of changes.

The mac tests need to be fixed, so that we can get a travis green. They have always been this way, but with Travis enabled now, we can't ignore it any more. One possibility could be to just do the compilation of the library, and run the tests after they are fixed to pass.

PR #75, opened 2015-01-08
Clean up handling of signgam

EdSchouten:
Hi there,

I am currently working on porting openlibm to an environment where the global variable signgam is not available. signgam is only part of the X/Open System Interface, not the C standard.

The attached changes reduce the coupling against signgam, making it easier to build a copy of openlibm that does not depend on it. It fixes tgammal() to not modify signgam. In addition to that, it replaces lgammal() by a simple wrapper around lgammal_r(), similar to lgammaf_r() and lgamma_r().

Ed

### DIFF
--- a/ld128/e_lgammal_r.c
+++ b/ld128/e_lgammal_r.c
@@ -16,7 +16,7 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-/*                                                      lgammal
+/*                                                      lgammal_r
  *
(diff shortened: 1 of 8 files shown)


ViralBShah:
These changes are all great. Once you're done with them, do let me know when we should tag a new release.

Also, just as an FYI, many of the complex functions, I recently stole from OpenBSD, since FreeBSD does not seem to have the complete set.

nalimilan:
@ViralBShah Speaking of a new release, do you think we could fix https://github.com/JuliaLang/openlibm/issues/55 soon?

PR #76, opened 2015-01-08
Change existing code to use CMPLX*() instead of cpack*() where possible.

EdSchouten:
I am currently working on building openlibm against stock copies of
<math.h>, instead of the openlibm.h header. It seems that a C compliant
<math.h> header can be used as a drop-in replacement for openlibm.h,
with the exception that it lacks cpack*().

In FreeBSD SVN r275819 I patched up the math library by replacing
cpack*() by CMPLX*(). That way many functions become less tied to the
intrinsics of the math library. Make the same change to openlibm.


Keno:
Thanks!

PR #77, opened 2015-01-09
Make openlibm less dependent on the host, but still add knobs to build against it

EdSchouten:
Hi there,

First of all, thank you guys for being so responsive to all of my previous pull requests. Well appreciated!

Attached is a somewhat largish patchset that essentially decouples OpenLibm from the host system. It patches the code up in such a way that it no longer includes `<fenv.h>`, `<complex.h>` and `<math.h>`. Instead, it will use `<openlibm_fenv.h>`, `<openlibm_complex.h>` and `<openlibm.h>`. `<openlibm_complex.h>` has been patched up to become feature-complete.

I've added some bits to the `<openlibm*.h>` to switch building against the host headers. I can use this switch to build the math library for the POSIX-like environment I'm working on.

Thoughts?

Ed

[Edit: Fix formatting for the headers, they didn't show up for me otherwise - @Keno]

### DIFF
--- a/amd64/fenv.c
+++ b/amd64/fenv.c
@@ -32,7 +32,7 @@
 #ifdef _WIN32
 #define __fenv_static
 #endif
-#include "fenv.h"
+#include "openlibm_fenv.h"
 
 #ifdef __GNUC_GNU_INLINE__
 #error "This file must be compiled with C99 'inline' semantics"
--- a/amd64/fenv.c
+++ b/amd64/fenv.c
@@ -32,7 +32,7 @@
 #ifdef _WIN32
 #define __fenv_static
 #endif
-#include "openlibm_fenv.h"
+#include <openlibm_fenv.h>
 
 #ifdef __GNUC_GNU_INLINE__
 #error "This file must be compiled with C99 'inline' semantics"
--- a/amd64/fenv.h
+++ b/amd64/fenv.h
@@ -29,8 +29,10 @@
 #ifndef	_FENV_H_
 #define	_FENV_H_
 
-#include "include/cdefs-compat.h"
-#include "include/types-compat.h"
+#include "cdefs-compat.h"
+#include "types-compat.h"
+
+#include "math_private.h"
 
 #ifndef	__fenv_static
 #define	__fenv_static	static
--- a/arm/fenv.c
+++ b/arm/fenv.c
@@ -27,7 +27,7 @@
  */
 
 #define	__fenv_static
-#include "fenv.h"
+#include <openlibm_fenv.h>
 
 #ifdef __GNUC_GNU_INLINE__
 #error "This file must be compiled with C99 'inline' semantics"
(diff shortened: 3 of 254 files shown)


tkelman:
We'll need to test this along with Julia's build system. Why did you choose to use the changed behavior by default? If for some reason this new changed self-contained mode doesn't work for Julia it would be less invasive for us to have the self-contained mode be the exceptional non-default case.

EdSchouten:
Hello Tony,

That's a very good question. I think it really depends on how we see OpenLibm as an individual project (not specifically focussing on Julia's use-case). As an outsider, my guess would be that there are two use-cases for OpenLibm:

1. As a separate library that can be used next to the system's math library.
2. As the actual math library of the system.

The question is which use-case we consider to be the default. If I check out the OpenLibm code and build it with the default options, I would personally expect it would be targeted at the first use-case. This is also how related libraries such as jemalloc (http://www.canonware.com/jemalloc/) work. It will simply install a jemalloc.h, containing functions like je_malloc(), je_free(), etc. People who want to use jemalloc or a math library as the core implementation of their operating system are likely the ones who mind getting their hands dirty.

That said, I personally don't care about the default behaviour. If you want, I can add the -DOPENLIBM_USE_HOST_*_H flags to the default CFLAGS.

Out of curiosity, how does the Julia language exactly use OpenLibm? Does the Julia runtime explicitly include <openlibm.h>, or is built against the system's <math.h>, hoping that OpenLibm provides the functions using the same ABI?

Ed

tkelman:
That's a very good point, mostly curious on your reasoning more than disagreeing with it. Also curious to find out more about the project you're working on.

By default Julia builds and calls into openlibm, but there's an option in our build system to link against system libm if you so choose. We've found a lot of variation in performance, accuracy, and correctness across different platforms' respective libm implementations, so nicer to have something uniform and predictable.

ViralBShah:
@EdSchouten Julia dlopen's and calls the functions in openlibm - so it does not care about openlibm.h.

This is the right set of patches - we really do not wish to depend on math.h. I do think that using the system provided headers is a good default - but I'd like to try that out on different systems for a while, before we make it default. Could you revert just that bit for now?

I'd like a separate PR for using system headers the default, where we can discuss the outcomes on various OSes, systems, and such.

tkelman:
FYI we also statically link `libopenlibm.a` into `julia.exe` on Windows.

EdSchouten:
Wow. I just discovered something interesting: it seems that OpenLibm already never depended on the `<fenv.h>` and `<math.h>` headers of the host system. It only included the `<complex.h>` header from the host system. Very confusing.

The only places where we included the host system's `<math.h>` is in functions in ld128/ that are not yet hooked up to the build yet. That said, including `<complex.h>` of the host system was quite harmless. This header essentially only defines _Complex_I, CMPLX*() and function prototypes. These have now been moved into `<openlibm_complex.h>`.

So to summarize: no behaviour has been changed substantially. The only thing I've done is made it more consistent (we no longer use the host system's `<complex.h>` ) and extended it to at least give the option to build against the host system's headers, which is needed for my specific use-case. The changes in this pull request could be merged as is.

I hope this explanation is clear enough. If not, be sure to get in touch.

Thanks,
Ed

PR #78, opened 2015-01-10
Cleanups

EdSchouten:
Hi there,

Attached is a small number of random, not so strongly related cleanup changes. A couple of these are actually needed for my use-case (fenv.h, OPENLIBM_ONLY_THREAD_SAFE). The others are just some cleanups for things I noticed along the way.

After these changes get merged, I no longer need to patch any OpenLibm source files. I only need tweak the Makefiles here and there.

Thanks,
Ed


EdSchouten:
Oh, wait. It turns out that our __strong_reference() macro only works if the target symbol has the same type as the source. Let me remove that part of the change for now.

ViralBShah:
Do you think the header files are all better off living in `include/`? Also, if we end up using the new header files, do they need to be installed as part of the `make install` step?

EdSchouten:
Thanks for merging!

Yes. I think having the files in include/ would be a bit better. I would even consider moving the internal headers that are stored into include/ into src/ instead. We currently use them the wrong way around in my opinion. The reason why I decided to move the headers into src/ was only because openlibm.h was already stored there.

As you mentioned, we should also add the install targets.

Would you be interested in addressing these issues, or shall I send you a patch?

Thanks,
Ed

ViralBShah:
You are right to do it the other way around. I realised it as I was looking at it. It would be great if you can send a PR.

Thanks again for all the patches. These tremendously increase the quality of openlibm.

PR #81, opened 2015-01-11
Rename openlibm.h to openlibm_math.h.

EdSchouten:
This is a bit more consistent with the naming of the other header files
(openlibm_complex.h and openlibm_fenv.h). Re-add an openlibm.h header
that includes all of the public headers as a shorthand.

Fix up all of the source files to include `<openlibm_math.h>` instead of
`<openlibm.h>`. While there, fix ordering of the includes.


ViralBShah:
Thanks. This is the best approach from our discussion on the topic.

PR #82, opened 2015-01-12
No stdio.h on Mini-OS, so use cdefs.h again there

talex5:


ViralBShah:
Sorry about this. Thanks for the fix. Since you should have commit access, feel free to push directly, or merge your PRs for such fixes.

Issue #79, opened 2015-01-10
cbrtl() is not being built, and POSIX conformance


EdSchouten:
It seems that s_cbrtl.c is not added to the list of source files in src/Make.files. It's explicitly commented out. Is this function still incomplete? If so, should we add a comment there? If not, shall we hook it up to the build?

ViralBShah:
I recollect having some issues with it earlier, which may no longer hold. If it builds, we should hook it up again.

EdSchouten:
Hi Viral,

It seems to build. I'll hook it up.

By the way, what are your thoughts in general on functions that are not part of the C standard/POSIX? In particular:

- gamma() (a misspelling of lgamma),
- finite() (a misspelling of isfinite),
- scalb() (a misspelling of scalbn),
- ...

I'd argue that these functions would normally be good to preserve if you have an existing commonly used implementation (FreeBSD's C library), but it would make little sense to provide those in newer or standalone implementations. I have these functions disabled in my local copy of OpenLibm, for example.

Maybe we could consider just removing them from the codebase and focussing on providing just the C/POSIX bits? If you want, I can do that at the same time.

ViralBShah:
How about having an `ifdef` so that the default is POSIX, but one can build with `-DLIBM_EXTENSIONS`? 

I think we may be using some of these in Julia. I certainly would want us to use the POSIX compliant bits only, but that would mean changing a few things on the julia side. I think POSIX compliance is the right thing to do - although I should confess that I don't know what all is part of the standard and what is not.

Once this is done, we will review the Julia usage, and remove the extensions.

ViralBShah:
BTW, if you know of any other libm test suites, they will be great to have.

EdSchouten:
Something like -DLIBM_EXTENSIONS sounds all right. I'll see what I can do.

ViralBShah:
@EdSchouten You are probably busy with other things, but would be great to have a PR to make the library POSIX compliant. Coming to think of it, we can avoid LIBM_EXTENSIONS altogether. I will fix whatever is needed on the Julia side, after the cbrtl and POSIX conformance changes, if you can provide those.

I am thinking of releasing openlibm 0.5 after these. Is there anything else we should do as part of the 0.5 release?

nalimilan:
@ViralBShah As I said, I think we should do something about https://github.com/JuliaLang/openlibm/issues/55 before a new release.

EdSchouten:
Hi there!

[...]

PR #85, opened 2015-02-12
Remove non-standard functions from OpenLibm

EdSchouten:
Hi Viral,

As mentioned in issue #79, I promised to send you a pull request to remove non-standard functions from OpenLibm, as they are only needed to keep existing implementations working. New implementations shouldn't offer them. This is how far I got.

Thanks,
Ed


ViralBShah:
Thank you!

simonbyrne:
Apparently julia still calls `significand` and `signficandf`: 
https://github.com/JuliaLang/julia/blob/1ec68b3173146178ec4f8340b7d2490d4164a4b8/base/math.jl#L135-L141

It would be fairly easy to implement these in Julia.

ViralBShah:
I hadn't checked which of these we were using, but the 0.4.1 release of openlibm that we use for now still has these, so julia is safe. For the 0.5 release of openlibm, we will have to update julia too.

nalimilan:
We're also going to have to bump the SOVERSION major number if we remove functions from the ABI. Better do it in this PR before we forget.

ViralBShah:
Already done in 763da446063bae2ccaff1893ba600739979c3ef3

tkelman:
We should put the single precision bessel functions back. Julia is using them and it doesn't hurt anything to have them in the library. The fact that a standard doesn't mention them is an omission in the standard. https://github.com/JuliaLang/julia/pull/13868

ViralBShah:
I have been putting such things in openspecfun. I guess we could perhaps have a flag in openlibm that allows extra stuff too. However, seems like @yuyichao has a pure julia solution, which is perhaps even better.

yuyichao:
The pure julia implementation is pretty close in performance but is statistically significantly slower than the openlibm or libm implementation (which are the same one AFAICT....) by ~30%. Using `muladd` bring that down to ~15%.

I think this is a pretty fun exercise and probably also a useful benchmark to have but I also agree with Tony that there's little point of writing this in julia if the C version is already working and widely tested. It might make sense to ship a julia version if it can beat the C performance or if it provides other benefit.

ViralBShah:
Agree. Just need a little bit more engg on the openlibm side for this. If we have an extensions flag in openlibm, we can even move rem_pio2 back from openspecfun back here.

tkelman:
Hasn't `USE_SYSTEM_LIBM` been passing? So glibc libm has these single precision bessel fcns? If that's the case I don't see why we would need an extensions flag. We can eventually move this kind of thing to a SpecialFunctions.jl with both Julia and C/Fortran implementations available, for example nolta's translation of Amos would help portability to systems that don't have fortran compilers. I should check whether MSVC supports these single precision Bessel fcns.

ViralBShah:
I believe they are not part of the (C?) spec, but perhaps commonly implemented? 

The flag should just be a couple of lines in Makefile. So, if we reinstate these, it should be straightforward to have a flag.

EdSchouten:
The `double` versions are part of the X/Open UNIX specification, not the C standard. Versions for `float` and `long double` are not part of any standard.

yuyichao:
> Versions for float and long double are not part of any standard.

Which is not a very convincing reason for not implementing them or even removing them from a existing library.

tkelman:
@ViralBShah no tagging openlibm until single-precision bessel gets put back

ViralBShah:
I guess all we need to do is revert https://github.com/JuliaLang/openlibm/commit/8c8693cf7962be7b0e53c491287aeb2fb13ae6f2

tkelman:
And there have been upstream bugfixes in https://github.com/freebsd/freebsd/commit/0563b7a42b5b5f9ab2af93daf62bfed48fb84e55 and several other commits, but if those bugs aren't causing test failures in Julia then they aren't necessarily release blocking.

ViralBShah:
4c8740ad589fd243eb2d3691f2c2395b505d713d fixes this. I'd still like to bring in the EXTENSIONS build flag, but this should be good for now to tag.

PR #86, opened 2015-02-20
Update to the latest ARM stuff from msun.

ViralBShah:
Many more tests pass now, but has to be compiled with -O0 to avoid
segfault in the rounding tests.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -45,7 +45,7 @@ ifeq ($(ARCH),mingw32)
 $(error "the mingw32 compiler you are using fails the openblas testsuite. please see the Julia README.windows.md document for a replacement")
 endif
 
-CFLAGS_add += -std=c99 -Wall -O1 -Wno-implicit-function-declaration -DASSEMBLER -D__BSD_VISIBLE -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(OPENLIBM_HOME)/ld80 -I$(OPENLIBM_HOME)/$(ARCH) -I$(OPENLIBM_HOME)/src
+CFLAGS_add += -std=c99 -Wall -O0 -Wno-implicit-function-declaration -DASSEMBLER -D__BSD_VISIBLE -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(OPENLIBM_HOME)/ld80 -I$(OPENLIBM_HOME)/$(ARCH) -I$(OPENLIBM_HOME)/src
 
 default: all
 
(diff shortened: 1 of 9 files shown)


Make.inc
```diff
@@ -43,7 +43,7 @@ ifeq ($(ARCH),mingw32)
 $(error "the mingw32 compiler you are using fails the openblas testsuite. please see the Julia README.windows.md document for a replacement")
 endif
 
-CFLAGS_add += -std=c99 -Wall -O3 -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(OPENLIBM_HOME)/ld80 -I$(OPENLIBM_HOME)/$(ARCH) -I$(OPENLIBM_HOME)/src -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration
+CFLAGS_add += -std=c99 -Wall -O1 -Wno-implicit-function-declaration -DASSEMBLER -D__BSD_VISIBLE -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(OPENLIBM_HOME)/ld80 -I$(OPENLIBM_HOME)/$(ARCH) -I$(OPENLIBM_HOME)/src
```
nalimilan: Why `-O1`? Is that required for ARM? Doesn't look like a good default for a performance-sensitive library.

ViralBShah: Actually it needs to be -O0. Tests are crashing at higher levels of optimization.

ViralBShah:
According to http://svnweb.freebsd.org/base/head/lib/msun/arm/Makefile.inc?view=log one has to do some strange things with the soft float ABI even for VFP, which is hard float.

ViralBShah:
With this patch, we go from all exception tests failing and a crash in the test suite, to passing with only 3 minor errors.

Make.inc
tkelman: can we make the change in optimization level arm-only?

ViralBShah: Yes, we could - but it will be kind of pointless. Better to just use the system libm till this is fixed.

nalimilan: I don't get it: do you want to merge this? This code is in openlibm, so people using this code are not using the system libm. Better keep `-O2` for i386/x86_64, where openlibm is used.

ViralBShah: I want to try a few more things, and then merge this with -O0 for ARM if I can't fix the segfault.

ViralBShah:
The arm branch needs more work. I have rebased, but it fails to find `sys/cdefs.h`.

ViralBShah:
This is folded into the `vs/msun` branch.

PR #90, opened 2015-02-27
Add support for cross compiling with `make ARCH=name`

### LINKED ISSUE #89: No real support for cross compiling
Cross compiling this is an enormous pain. It should be easier. Currently if I want to cross compile to i386 I need to run: `make LDFLAGS="-m32 -march=i386" SFLAGS="-m32 -march=i386" CFLAGS="-m32 -march=i386" ARCH=i386` which is really stupid.

scialex:
This is rather basic. A clean is needed to get rid of artifacts from
previous runs with other architectures.

Note this requires PR #88 to work because without it x86 is broken.

Closes #89

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -38,7 +38,9 @@ CC = gcc
 CFLAGS_add += -fno-gnu89-inline
 endif
 
-ARCH := $(shell $(CC) -dumpmachine | sed "s/\([^-]*\).*$$/\1/")
+ARCH ?= $(shell $(CC) -dumpmachine | sed "s/\([^-]*\).*$$/\1/")
+ARCH_ORIGIN := $(origin ARCH)
+
 ifeq ($(ARCH),mingw32)
 $(error "the mingw32 compiler you are using fails the openblas testsuite. please see the Julia README.windows.md document for a replacement")
 endif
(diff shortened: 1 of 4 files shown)


scialex:
I'm not really able to test on arm ATM, if one of you could that would be great.

ViralBShah:
This is great. Could you add some documentation to README.md about cross compiling, with some examples?

Should we be autoconf-ing openlibm?

scialex:
Done.

>Should we be autoconf-ing openlibm?

I don't really know anything about autoconf TBH. It might be a good idea but I don't know what the process for it would be.

tkelman:
> Should we be autoconf-ing openlibm?

No. We should be cmake-ing openlibm.

ViralBShah:
I guess cmake is the new autoconf. Need to figure it out sometime.

kmsquire:
> > Should we be autoconf-ing openlibm?

> No. We should be cmake-ing openlibm.

While I've come to rather like cmake while working with it in other projects, I thought that one of its shortcomings was [weak support for cross compiling](https://github.com/JuliaLang/julia/issues/10010#issuecomment-72482499) (which this PR was addressing).

tkelman:
I'm not sure if "weak support" is quite the right phrasing, vs making things a bit more cumbersome than autotools does. Note that the method used in this PR, and referred to for base Julia in that issue, isn't a conventional cross-compilation using a host prefix which works ok in cmake (you just need a lot of flags or a prewritten toolchain file for it). This `-m32` method is really multilib compilation. In either autotools or cmake I think you would do this the same way, via cflags.

PR #97, opened 2015-08-13
Update Make.inc to detect OpenBSD

ziotom78:
With this patch I have been able to successfully compile Openlibm on OpenBSD 5.7 (x86_64) using the following command:

    gmake GCC=egcc GXX=eg++ USEGCC=1 ARCH=x86_64

I had to install gcc 4.9 from ports, as Openlibm does not compile with the default compiler shipped with OpenBSD 5.7, which is gcc 4.2.1. Also, gnumake must be installed for this to work.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -103,6 +103,12 @@ SONAME_FLAG = -soname
 CFLAGS_add+=-fPIC
 endif
 
+ifeq ($(OS), OpenBSD)
+SHLIB_EXT = so
+SONAME_FLAG = -soname
+CFLAGS_add+=-fPIC
+endif
+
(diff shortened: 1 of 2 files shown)


ViralBShah:
Could you add the instructions to the build instructions to README.md as well?

ziotom78:
I have updated `README.md`, however I am not sure of the quality of my additions. I just added a bullet point under the section "Building instructions", but my text looks a bit out of place (all the other points are quite general, while mine is very specific). Perhaps moving it to a dedicated subsection (maybe under "Platform support") would be better.

ViralBShah:
Thanks. Will clean up the README further later. This is good for now.

PR #98, opened 2015-10-26
Fix missing buildflags

ginggs:
Debian's blhc (build log hardening check) reports dpkg-buildflags-missing CPPFLAGS 48 (of 223), CFLAGS 2 (of 177), LDFLAGS 2 (of 3) missing

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -53,7 +53,7 @@ default: all
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(CFLAGS_add) -c $< -o $@
 
 %.S.o: %.S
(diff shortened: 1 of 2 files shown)


Keno:
Thanks!

nalimilan:
Wait, why are you passing `CPPFLAGS` when compiling C files? I would say either  `CFLAGS`, `CPPFLAGS` or  `FFLAGS` should be passed, but never two at the same time.

yuyichao:
`CPPFLAGS` is for the [c/c++ pre-processor](https://en.wikipedia.org/wiki/CFLAGS). `CXXFLAGS` is for c++.

nalimilan:
Sorry, I always confuse them. Carry on...

PR #99, opened 2015-10-26
Fix underlinking

ginggs:
libopenlibm should be linked to libm to resolve the symbols logl, exp and powl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ libopenlibm.$(SHLIB_EXT): $(OBJS)
 ifeq ($(OS),WINNT)
 	$(CC) -shared $(OBJS) $(LDFLAGS) $(LDFLAGS_add) -Wl,$(SONAME_FLAG),libopenlibm.$(SHLIB_EXT) -o libopenlibm.$(SHLIB_EXT)
 else
-	$(CC) -shared $(OBJS) $(LDFLAGS) $(LDFLAGS_add) -Wl,$(SONAME_FLAG),libopenlibm.$(SHLIB_EXT).$(SOMAJOR) -o libopenlibm.$(SHLIB_EXT).$(SOMAJOR).$(SOMINOR)
+	$(CC) -shared $(OBJS) $(LDFLAGS) $(LDFLAGS_add) -Wl,$(SONAME_FLAG),libopenlibm.$(SHLIB_EXT).$(SOMAJOR) -lm -o libopenlibm.$(SHLIB_EXT).$(SOMAJOR).$(SOMINOR)
 	@-ln -sf libopenlibm.$(SHLIB_EXT).$(SOMAJOR).$(SOMINOR) libopenlibm.$(SHLIB_EXT).$(SOMAJOR)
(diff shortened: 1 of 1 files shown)


ginggs:
The following warnings in amd64 and i386 builds were fixed by explicitly linking libm:

dpkg-shlibdeps: warning: symbol logl used by debian/libopenlibm1/usr/lib/i386-linux-gnu/libopenlibm.so.1.0 found in none of the libraries
dpkg-shlibdeps: warning: symbol expl used by debian/libopenlibm1/usr/lib/i386-linux-gnu/libopenlibm.so.1.0 found in none of the libraries
dpkg-shlibdeps: warning: symbol powl used by debian/libopenlibm1/usr/lib/i386-linux-gnu/libopenlibm.so.1.0 found in none of the libraries

See: https://buildd.debian.org/status/fetch.php?pkg=openlibm&arch=i386&ver=0.4.1%2bdfsg-1&stamp=1430399652

ginggs:
I don't know if this is needed for WINNT as well.

ginggs:
What is worrying is that __fpgetprec is still unresolved on i386:

dpkg-shlibdeps: warning: symbol __fpgetprec used by debian/libopenlibm1/usr/lib/i386-linux-gnu/libopenlibm.so.1.0 found in none of the libraries

According to amd64/bsd_ieeefp.h (included from 387/bsd_ieeefp.h) this is an inline function:
static __inline__ fp_prec_t __fpgetprec(void)

This might have something to do with #55.

ViralBShah:
I would really prefer not linking to libm, and instead getting those functions compiling in openlibm. Even though more work, it is worth getting it right.

ginggs:
I've created PR #103 to fix the unresolved symbol __fpgetprec on i386.

ViralBShah:
So what should we do here for now? Is there risk that libm versions will get called instead of openlibm?

petercolberg:
What is the current status of openlibm with regard to API compatibility with libm? Does openlibm implement all functions in libm? If not, I suggest disabling those `long double` functions with unresolved symbols, until `powl` etc are implemented in openlibm.

petercolberg:
For further context, please also see this [discussion on the Debian Julia list](https://lists.alioth.debian.org/pipermail/pkg-julia-devel/Week-of-Mon-20151109/000809.html).

petercolberg:
I am glad to note that this issue has already been solved in HEAD.

The underlinking issue applies only to version 0.4.1, which does not yet contain definitions of `expl`, `logl`, and `powl`; hence the following error when linking the tests against the shared library:
```
make[2]: Entering directory '/build/openlibm-0.4.1+dfsg/test'
gcc -D_FORTIFY_SOURCE=2 -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wl,-z,relro -g test-double.c 
../libopenlibm.so: undefined reference to `powl'
../libopenlibm.so: undefined reference to `logl'
../libopenlibm.so: undefined reference to `expl'
```

ViralBShah:
Can we close this PR?

PR #103, opened 2015-10-28
Add missing include to inline __fpgetprec on i386

ginggs:
This fixes the unresolved symbol __fpgetprec in libopenlibm.so.1.0 on i386


ginggs:
I still want to check if this has anything to do with #55 .

PR #104, opened 2015-11-12
Test shared library

petercolberg:
This patch adds a make check target that builds and runs the tests using the shared library.

This is a prerequisite for resolving issue #99.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -21,5 +21,4 @@ arm32)
 	;;
 esac
 
-$LOADER ./test/test-double
-$LOADER ./test/test-float
+$LOADER make check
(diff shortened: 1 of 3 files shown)


tkelman:
lgtm

petercolberg:
Both builds on Travis succeeded. Did you intend to comment on a different pull request?

ViralBShah:
The failures are only on mac - perhaps because of the use of clang.

tkelman:
Travis isn't running on osx here, ref https://github.com/JuliaLang/openlibm/commit/5473e8b27deb55986bc3456bc3f574cdebf174c5

ViralBShah:
I just realised that. We probably need to import a few more clang fixes from msun from over the last few months, and that may make things pass all the way, or at least more.

PR #105, opened 2015-11-13
[WIP] More makefile improvements

petercolberg:
I am going over the patches in the openlibm Debian package to see which ones are useful to integrate into upstream openlibm.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -45,7 +45,11 @@ ifeq ($(ARCH),mingw32)
 $(error "the mingw32 compiler you are using fails the openblas testsuite. please see the Julia README.windows.md document for a replacement")
 endif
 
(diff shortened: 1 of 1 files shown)


petercolberg:
Sorry for not being clear, WIP stands for work in progress…

petercolberg:
Could you undo the merge?

ViralBShah:
Of course. Didn't see that. I think I can only revert it for now. I have also sent you commit access so you could do it as well.

petercolberg:
You can revert the merge using
```
git push origin -f 4c8740a:master
```
as long as noone has fetched so far, which is reasonable to assume but should remain the absolute exception.

ViralBShah:
It's a protected branch, it turns out. If you don't mind, we can revert, or do a new PR for the remaining things.

petercolberg:
I noticed the protected branch. That is good practice anyway.

Then please leave it as is. The commit itself is good, I was just going to add more related commits to the pull request.

nalimilan:
Are we sure this `-O3` flag is really useful as a default? My understanding is that it's very rarely a good idea, and should only be applied when benchmarking shows a real improvement.

nalimilan:
Also, @ViralBShah, could you add @petercolberg and @xyproto (for Arch) to the packagers team? I just discovered this allows CCing all team members by writing `@JuliaLang/distro-packagers `.

ViralBShah:
I think `-O2` should be good.

ViralBShah:
Done.

PR #106, opened 2015-11-13
Set default optimization flag to -O2

petercolberg:
This sets the default optimization level to -O2 as discussed in issue #105. The optimization flag may be overriden with CFLAGS, which the canonical way in distributions such as Debian.

On a general note, please allow sufficient time to the openlibm team members to review before merging.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -22,3 +22,5 @@ arm32)
 esac
 
 $LOADER make check
+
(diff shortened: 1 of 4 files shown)


ViralBShah:
All look good.

ViralBShah:
Shall we merge this?

PR #107, opened 2015-11-28
WIP: Updating to the latest msun (Do not merge)

ViralBShah:


src/e_lgamma.c
```diff
@@ -21,16 +21,19 @@
...
+#include <openlibm_math.h>
 #include "math_private.h"
 
-DLLEXPORT double
+extern int signgam;
+
+double
 __ieee754_lgamma(double x)
 {
-#ifdef OPENLIBM_ONLY_THREAD_SAFE
```
tkelman: ref https://github.com/JuliaLang/openlibm/commit/c253db68caea45ee56b7d35e93bb2945b98ddf5b from #78

ViralBShah: I still think that we need the `extern int signgam` if not in the THREAD_SAFE compile mode. All these patches should perhaps be external patches, so that we can sync with msun upstream and then apply our patches.

tkelman: Yeah, figuring out a way to automate the update-and-rebase process would be a really good idea. Maybe something with subtree or sparse checkout could preserve the git history on just the libm parts of the upstream source?

PR #110, opened 2016-01-19
FreeBSD has different pkg-config dir: /usr/local/libdata/

iblis17:
Also, let the `prefix` be configurable from outside(e.g `make prefix=/path`) .

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -12,6 +12,12 @@ bindir = $(prefix)/bin
 libdir = $(prefix)/lib
 includedir = $(prefix)/include
 
+ifeq ($(OS), FreeBSD)
+  pkgconfigdir = $(prefix)/libdata/pkgconfig
+else
+  pkgconfigdir = $(libdir)/pkgconfig
+endif
+
 USEGCC = 1
 USECLANG = 0
 
@@ -91,7 +97,7 @@ ifeq ($(OS), WINNT)
   SONAME_FLAG = -soname
   override CFLAGS_add += -nodefaultlibs
   shlibdir = $(bindir)
- else
+else
   ifeq ($(OS), Darwin)
     SHLIB_EXT = dylib
     SONAME_FLAG = -install_name
--- a/Makefile
+++ b/Makefile
@@ -62,9 +62,9 @@ openlibm.pc: openlibm.pc.in Make.inc Makefile
 
 install: all openlibm.pc
 	mkdir -p $(DESTDIR)$(shlibdir)
-	mkdir -p $(DESTDIR)$(libdir)/pkgconfig
+	mkdir -p $(DESTDIR)$(pkgconfigdir)
 	mkdir -p $(DESTDIR)$(includedir)/openlibm
 	cp -f -a libopenlibm.$(SHLIB_EXT)* $(DESTDIR)$(shlibdir)/
 	cp -f -a libopenlibm.a $(DESTDIR)$(libdir)/
 	cp -f -a include/openlibm*.h $(DESTDIR)$(includedir)/
-	cp -f -a openlibm.pc $(DESTDIR)$(libdir)/pkgconfig/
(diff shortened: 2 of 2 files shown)


Make.inc
```diff
@@ -7,9 +7,10 @@ VERSION = 0.5-dev
 SOMAJOR = 2
 SOMINOR = 1
 DESTDIR =
-prefix = /usr/local
+prefix ?= /usr/local
```
nalimilan: This isn't needed AFAIK. All variables can be overridden already.

nalimilan:
Out of curiosity, do you have any references about `libdatadir`? I've never seen other libraries use this variable in their makefiles, and it's not listed e.g. in the GNU make standard variables. So I wonder how FreeBSD handles all those programs that don't support it.

iblis17:
@nalimilan Oh...  `libdatadir` is just my naming, i'm willing to change it if you wish.
Some project use `pkgconfigdir` or whatever you can image.

We can find an example here: [libzmq](https://github.com/zeromq/libzmq/blob/1a502cc190228aef226aadb16808def222c4e09f/configure.ac#L623-L628) have `pkgconfigdir` configurable.
And there is a [freebsd ports](https://github.com/freebsd/freebsd-ports/blob/master/net/libzmq4/Makefile#L35) for it, set it correctly.

For the project which do not support it yet, we are encouraged to report to upstream. Or, we will make a patch for `Makefile`, uploud to freebsd ports. This patch will be apply during installation via ports.

Make.inc
iblis17: Oh. ok i will remove it :smile:

nalimilan:
OK. `pkgconfigdir` looks more common (there are autoconf macros about it), and I find it more explicit.

nalimilan:
(But, really, this sounds like a heroic stance from FreeBSD people to decide to patch all upstream projects for a detail like that... :-)

iblis17:
(Ha. BTW. There is still lots of project using bash script with `#!/bin/bash`, and this is not portable for bsd... I always do PR for this: "hey! our bash is here(`/usr/local/bin/bash`) ... `#!/usr/env bash` would be portable.")

iblis17:
renaming done.

Make.inc
```diff
@@ -10,6 +10,7 @@ DESTDIR =
 prefix = /usr/local
 bindir = $(prefix)/bin
 libdir = $(prefix)/lib
+pkgconfigdir = $(prefix)/libdata/pkgconfig
```
nalimilan: Please keep the default as `$(libdir)/pkgconfig`.

Make.inc
```diff
@@ -103,6 +104,12 @@ ifeq ($(OS), WINNT)
   shlibdir = $(libdir)
 endif
 
+ifeq ($(OS), FreeBSD)
```
nalimilan: You should move this code above, and use it instead of the unconditional definition. Also, use `$(libdir)` as I said above, since people may override it.

Make.inc
iblis17: Oops. sorry for that

Make.inc
iblis17: hmm...  you mean merge this `if` block in to the previous `if` block ?

nalimilan: No, I mean move this to the first place where you define `pkgconfigdir`.

iblis17: Oh, place it after line 15?
and... i try to understand your statement: "use it instead of the unconditional definition".
Would you explain more? I cannot follow your thinking right now.

nalimilan: Instead of line 13.

nalimilan:
Thanks, LGTM.

@ViralBShah Good to go?

iblis17:
@nalimilan Thanks for your patient reviewing.  :)

tkelman:
This looks okay to me

PR #111, opened 2016-01-20
use `-O0` to disable optimization in testing

wdv4758h:


nalimilan:
Not sure that's a good idea. It's essential to run the tests with the settings actual users are going to use. `-O0` could be an additional build in the Travis CI matrix (just like using clang in addition to gcc).

nalimilan:
Ref: https://github.com/JuliaLang/openlibm/issues/10

wdv4758h:
I think we can leave it in Travis CI matrix (clang, `-O0`, ... etc), but we still need `-fno-builtin` to avoid problem like #10 (maybe another PR).

PR #116, opened 2016-02-23
work around optimisation bugs for rint

simonbyrne:
the `rint` functions (`lrint`, `nearbyint`) etc cause problems on ARM and power when compiled with optimisations. This disables optimisations for those functions, and enables them for all others. Also reenables a disable test.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -49,6 +49,13 @@ CFLAGS_add += -std=c99 -Wall -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(O
 
 default: all
 
+# *int / *intf need to be built with -O0
+src/%int.c.o: src/%int.c
+	$(CC) $(CPPFLAGS) -O0 $(CFLAGS_add) -c $< -o $@
+
+src/%intf.c.o: src/%intf.c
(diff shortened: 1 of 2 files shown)


Make.inc
```diff
@@ -49,6 +49,13 @@ CFLAGS_add += -std=c99 -Wall -I$(OPENLIBM_HOME) -I$(OPENLIBM_HOME)/include -I$(O
 
 default: all
 
+# *int / *intf need to be built with -O0
```
tkelman: this is true even on x86?

EdSchouten:
What does the optimizer do that causes these tests to fail? Is it a bug in the compiler, or is it a bug in the code?

simonbyrne:
> What does the optimizer do that causes these tests to fail? Is it a bug
in the compiler, or is it a bug in the code?

I'm not sure exactly, though I suspect it may be due to gcc's poor fenv
support.

Make.inc
ViralBShah: Yes. See the failing test for `nearbyint` below that I had commented out a while back, and passes as a result of this.

ViralBShah:
Verified on arm.

ViralBShah:
@edelsohn mentions that gcc 5 implements much of this right for most targets.

PR #118, opened 2016-02-24
New PR to sync up with msun

ViralBShah:
In order to keep this manageable in the long term, we need to import code from msun and openbsd, and make no modifications to it. Ideally we should have our own headers, and maintain any other changes to sources as patches - so that we can frequently stay updated.


ViralBShah:
With this PR, much of the stuff compiles without modification of the upstream sources, but this error has me stumped. Any ideas?

````
clang  -O2 -fno-builtin -std=c99 -Wall -I/Users/viral/repos/openlibm -I/Users/viral/repos/openlibm/include -I/Users/viral/repos/openlibm/ld80 -I/Users/viral/repos/openlibm/amd64 -I/Users/viral/repos/openlibm/src -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -fPIC -c src/s_csqrt.c -o src/s_csqrt.c.o
src/s_csqrt.c:65:17: error: use of undeclared identifier 'INFINITY'
                return (CMPLX(INFINITY, b));
                              ^
/Users/viral/repos/openlibm/include/openlibm_complex.h:57:41: note: expanded
      from macro 'CMPLX'
#  define CMPLX(x, y) ((double complex){x, y})
                                        ^
1 error generated.
make: *** [src/s_csqrt.c.o] Error 1
````

ViralBShah:
The above error is fixed. The tests are passing 100% on linux, but on os x, we have tons on `Invalid operation` errors. @simonbyrne You seem to have figured out how to fix them on powerpc. Would be great if you could take a look.

iblis17:
Hi @ViralBShah 
About the testing related issue, i will highlight [this comment in #10](https://github.com/JuliaLang/openlibm/issues/10#issuecomment-173203675).

I hope that the `test/Makefile` can add the flag `-no-bulitin`.

ViralBShah:
And added to the tests too - f7a18b519dc6ba654fce73fdfb554d7781c18733

PR #124, opened 2016-02-27
Upstream local modifications to openlibm.

EdSchouten:
I currently have a small number of local changes that I need to apply on top of openlibm to properly integrate it into CloudABI's C library. It would be nice if I could get rid of them. ;-)


nalimilan:
I must say this sentence is confusing to me  "Clang doesn't dislike it if you bitshift negative numbers, as the behaviour of it is undefined." :-)

EdSchouten:
Oh, sorry about that. Fixed!

Issue #117, opened 2016-02-23
Tag a new release - 0.5


ViralBShah:
With the ppc support, I think we should tag a new release - 0.5. Is there anything else we need to do before we tag?

tkelman:
We should probably run julia's tests against it on a variety of platforms.

ViralBShah:
What's the best way to do so? The changes here are minor and there are no api changes - so I think this is pretty safe.

StefanKarpinski:
Is it likely that there will be and API changes in the future? If not, then a 1.0 release may be warranted. Bugfixes would go into 1.0.x releases; new exported functions would go in 1.x releases, and any breaking API changes (including deleting exported functions) would require a 2.0 release.

tkelman:
> What's the best way to do so?

Prepare the PR that changes the `openlibm.version` and checksum files first, run it through CI before doing the tag here?

> Is it likely that there will be and API changes in the future?

Not necessarily API changes, but in terms of code #108 will be pretty big and still needs some cleanup. I also think we should look carefully at what musl has done with their libm since they started from pretty much the same code bases we did, but have been better at maintaining and pulling in bug fixes.

ViralBShah:
Can you point to the musl implementation? The issue we have faced is that we never automated the migration from msun. It needs a bunch of sed scripts really, and unfortunately I didn't save that in #108. I'll have to do it once more carefully.

ViralBShah:
By definition, there should be no major API changes, since the goal is to be a libm replacement. I would like to sort out #108 and keeping up with msun before tagging 1.0.

tkelman:
http://git.musl-libc.org/cgit/musl/tree/src/math - looks like they don't have powerpc support at the moment though

ViralBShah:
That code looks quite clean! Do they maintain a record of what all they pull in from msun and how frequently?

tkelman:
http://git.musl-libc.org/cgit/musl/log/src/math

ViralBShah:
I meant tracking the msun logs. Looks like it started from msun, but has significantly changed, and ongoing fixes are manually tracked. I'll try this out to see if there is merit in following musl instead of msun before refreshing msun.

ViralBShah:
Detailed notes on the musl approach - http://nsz.repo.hu/libm/ 

Many functions have been completely rewritten in musl.

tkelman:
[...]

PR #125, opened 2016-02-28
Fix various warnings

nalimilan:
Cf. https://github.com/JuliaLang/openlibm/issues/117.


src/fpmath.h
```diff
@@ -41,30 +41,23 @@
...
-
 /* Definitions provided directly by GCC and Clang. */
-#define _LITTLE_ENDIAN    __ORDER_LITTLE_ENDIAN__
-#define _BIG_ENDIAN       __ORDER_BIG_ENDIAN__
-#define _PDP_ENDIAN       __ORDER_PDP_ENDIAN__
-#define _BYTE_ORDER       __BYTE_ORDER__
+#if !(defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && defined(__ORDER_BIG_ENDIAN__))
 
-#elif defined(__GLIBC__)
+#if defined(__GLIBC__)
```
EdSchouten: Instead of leaving all of this code there, would it be worth just throwing it away and seeing what breaks? It may be the case that Windows is the only platform that doesn't define these. All the other systems use a compiler that provides support for these definitions.

nalimilan:
Test failure happens since https://github.com/JuliaLang/openlibm/commit/f7a18b519dc6ba654fce73fdfb554d7781c18733.

src/fpmath.h
```diff
@@ -109,14 +98,14 @@ union IEEEf2bits {
 union IEEEd2bits {
 	double	d;
 	struct {
-#if _BYTE_ORDER == _LITTLE_ENDIAN
-#if _IEEE_WORD_ORDER == _LITTLE_ENDIAN
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#if _IEEE_WORD_ORDER == __ORDER_LITTLE_ENDIAN__
```
EdSchouten: At least GCC provides __FLOAT_WORD_ORDER__ for this purpose. Would it make sense to also use this name directly instead of using _IEEE_WORD_ORDER?

src/fpmath.h
nalimilan: Given that I have no way to check other compilers, I'd rather keep these as it's not much code and it's pretty clear. Do you have any reference about what other compilers support? The only thing I have found is https://sourceforge.net/p/predef/wiki/Endianness/, which doesn't seem to say so.

src/fpmath.h
nalimilan: Indeed. Will fix.

src/fpmath.h
```diff
@@ -72,26 +65,22 @@
...
+#define __BYTE_ORDER__           __ORDER_LITTLE_ENDIAN__
 
 #endif
 
-#ifndef _IEEE_WORD_ORDER
-#define	_IEEE_WORD_ORDER	_BYTE_ORDER
+#endif /* __BYTE_ORDER__, __ORDER_LITTLE_ENDIAN__ and __ORDER_BIG_ENDIAN__ */
+
+#ifndef __FLOAT_WORD_ORDER__
+#define	__FLOAT_WORD_ORDER__     __BYTE_ORDER__
```
EdSchouten: This is missing a space between `#define` and `__FLOAT_WORD_ORDER__`, right?

src/fpmath.h
EdSchouten: Speaking on behalf of FreeBSD, the `__FreeBSD__` block can safely be removed.

amd64/bsd_asm.h
```diff
@@ -40,7 +40,7 @@
 #include "../i387/osx_asm.h"
 #define CNAME(x) EXT(x)
 #else
-#include "bsd_cdefs.h"
+#include <sys/cdefs.h>
```
tkelman: musl specifically says "you should not be including this" http://wiki.musl-libc.org/wiki/FAQ#Q:_I.27m_trying_to_compile_something_against_musl_and_I_get_error_messages_about_sys.2Fcdefs.h

nalimilan: That's what I feared. I hate C. You can't rely on these macros to be defined as the header is private, but if you define them you get a redefinition warning.

Let's go back to only checking for `_SYS_CDEFS_H` and `_SYS_CDEFS_H`.

yuyichao: You can just rename them.

src/fpmath.h
nalimilan: Good catch, the file originally contained a tab. Fixed.

amd64/bsd_asm.h
nalimilan: That's another solution, but that would require a lot of cleanup since there are a lot of macros defined in these files (most of which I hope aren't used).

nalimilan:
I've pushed a new commit removing all unused macros from `bsd_cdefs.h`, and I must say I quite like it (about -1000 LOC). Even without renaming them, this shuts down the warnings on Linux.

BTW, it seems to me that in `math_private.h`, `#ifdef __GNUCLIKE_ASM` is always false, as only `bsd_cdefs.h` defines it, and it's not included AFAICT. Am I right?

ViralBShah:
I have reverted the -fno-builtins for now to get CI running again, and will turn it on more systematically later.

ViralBShah:
If we can rebase this on master, we may be able to get a green.

nalimilan:
Cool, I've just rebased.

Note that I've kept `bsd_cdefs.h` and `cdefs-compat.h` separate even if they are much smaller now, because merging them triggers errors I wasn't able to debug. Since you know the codebase much better, you might be able to clean this a bit more.

simonbyrne:
Thanks, that fixes the warnings on Power.

ViralBShah:
Cc @edelsohn

PR #126, opened 2016-03-03
More header fixes

nalimilan:
The first commit fixes a warning on Fedora rawhide with GCC6. But we should check that this works on Windows.

The second commit is just a cleanup.

### DIFF
--- a/i387/fenv.c
+++ b/i387/fenv.c
@@ -26,10 +26,10 @@
  * $FreeBSD: src/lib/msun/i387/fenv.c,v 1.8 2011/10/21 06:25:31 das Exp $
  */
 
-#include <cdefs-compat.h>
-#include <types-compat.h>
-#include <math_private.h>
-#include <i387/bsd_npx.h>
+#include "cdefs-compat.h"
+#include "types-compat.h"
+#include "math_private.h"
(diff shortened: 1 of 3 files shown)


nalimilan:
I've enabled OS X on Travis CI in a branch, and the build works (though tests fail): https://travis-ci.org/JuliaLang/openlibm/jobs/113449199

But there's still another warning about redefining `__pure2`, which is already defined by `sys/cdefs.h`. I'm not sure what we should do here: define it only if not yet defined, so that we trust the system to do what's right? Or call it `__openlibm_pure`?

ViralBShah:
I would prefer to do the conditional define. Changing these things more means more difficulty in importing upstream changes.

ViralBShah:
The failing cases are all 1 ulp and some FP exception stuff. I would say that if no new failures arise, we should be ok.

ViralBShah:
If we can address the `__pure2` stuff, let's get this in for the 0.5 release. Nice to have but not essential.

nalimilan:
I'm preparing a fix.

nalimilan:
With this small patch it should be good to go. I agree the errors can wait for another release since they aren't regressions.

Issue #127, opened 2016-03-12
libopenlibm.so.2.1 exported symbols


ginggs:
I am analyzing the symbols exported by the openlibm 0.5.0 library (libopenlibm.so.2.1).
In Debian, we use [symbols files](https://wiki.debian.org/UsingSymbolsFiles) to track in which versions of libraries new symbols appear and old symbols disappear.

In openlibm 0.5.0, we gained some symbols, like:

```
__scan_nan@Base
cacos@Base
cacosf@Base
```

...and we lost some symbols, like:

```
__fedisableexcept@Base
_scan_nan@Base
gamma@Base
```

This is normal for a library where the SONAME has been bumped.

I did notice that there were several symbols that were missing from the **powerpc** build, for example:

```
_ItL_aT@Base
_ItL_atanhi@Base
_ItL_atanlo@Base
```

...from 0.4 which are still present in 0.5.0 on **amd64** and **i386**, and several symbols names ending in 'l' from 0.4 and 0.5.0, for example:

```
acoshl@Base
acosl@Base
fabsl@Base
```

...which are present on **amd64** and **i386**.

There were also some symbols present on **powerpc** that were only present on **i386** and not present on **amd64**, for example:

```
feclearexcept@Base 0.4
fegetexceptflag@Base 0.4
fegetround@Base
```

I attach a symbols file that covers the architectures on which I have built openlibm so far; **amd64**, **i386**, **powerpc** and **ppc64el**.  I hope the format of the file is easy enough to follow.
The number at the end of a line is the version where the symbol first appeared.
Lines beginning '#MISSING:' are comments which indicate symbols that have been dropped.

Example 1: the following symbol only appears on **i386**, **kfreebsd-i386** and **hurd-i386**:

```
(arch=i386 kfreebsd-i386 hurd-i386)__has_sse@Base 0.4
```

Example 2: the following symbol appears on all architectures except **powerpc** and **ppc64el**:

```
 (arch=!powerpc !ppc64el)__kernel_sinl@Base 0.4
```

Any comments?

[libopenlibm2.symbols.txt](https://github.com/JuliaLang/openlibm/files/170376/libopenlibm2.symbols.txt)

ViralBShah:
I have no idea what to do here. Can you suggest?

petercolberg:
[...]

PR #128, opened 2016-03-13
Fix Clang warnings on Windows

Keno:
- Align DLLEXPORT in definitions and declations. There is still a few
  cases left, where the declation in the compiler's complex.h disagrees
  with the implementation here. For now we can't do anything about that,
  but maybe should be revisited in the future.
- Fix the syntax on an .ascii directive that gcc accepted mistakingly, but
  clang does not.


tkelman:
We should probably be using `OLM_DLLEXPORT` here to avoid collisions.

ViralBShah:
Does this warrant a new point release when merged?

nalimilan:
> Does this warrant a new point release when merged?

Maybe check whether we want to do something about https://github.com/JuliaLang/openlibm/issues/127 first?

Keno:
> We should probably be using OLM_DLLEXPORT here to avoid collisions.

Yes, I agree. Will do the rename.

PR #129, opened 2016-04-03
Fix building tests on GNU/kFreeBSD and GNU/Hurd

ginggs:
Since 4c8740ad589fd243eb2d3691f2c2395b505d713 -rpath is only set for $(OS) = Linux.
If setting -rpath only breaks OS X, then it may be better to set -rpath everywhere except where $(OS) = Darwin.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -3,7 +3,7 @@ include ../Make.inc
 
 # Set rpath of tests to builddir for loading shared library
 OPENLIBM_LIB = -L.. -lopenlibm
-ifeq ($(OS),Linux)
+ifneq ($(OS), Darwin)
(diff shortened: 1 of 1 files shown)


tkelman:
Don't think Windows needs rpath either

ginggs:
@tkelman 
> Don't think Windows needs rpath either
I agree, but does having `-rpath` set break the Windows build?  I haven't tested that it breaks an OS X build either.

tkelman:
Would have to test whether that flag is accepted.

CDLuminate:
I guess this is already fixed? https://github.com/JuliaLang/openlibm/blob/master/test/Makefile#L6-L10

ginggs:
Fixed in #167.

PR #131, opened 2016-04-08
WIP: Enable build on aarch64

ginggs:
This has been made much easier since recent commits by the addition of src/aarch64_fpmath.h and enabling builds for powerpc.

Mostly copied verbatim from Edmund Grimley Evans' patch in [Debian bug #803206](https://bugs.debian.org/803206).

So far, it builds and passes tests.  I haven't tried building Julia with it yet,

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -73,6 +73,9 @@ REAL_ARCH := $(ARCH)
 ifeq ($(findstring arm,$(ARCH)),arm)
 override ARCH := arm
 endif
+ifeq ($(ARCH),aarch64)
+override ARCH := arm
+endif
 ifeq ($(findstring powerpc,$(ARCH)),powerpc)
 override ARCH := powerpc
 endif
--- a/bsdsrc/b_tgamma.c
+++ b/bsdsrc/b_tgamma.c
@@ -128,16 +128,16 @@ tgamma(x)
 {
 	struct Double u;
 
-	if (x >= 6) {
+	if (isgreaterequal(x, 6)) {
 		if(x > 171.63)
 			return (x / zero);
 		u = large_gam(x);
 		return(__exp__D(u.a, u.b));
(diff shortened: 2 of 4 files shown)


yuyichao:
Nice, I'll check if it passes julia tests =)

yuyichao:
Julia build fine and `all` tests passed!

Julia version I run the tests on:

```jl
Julia Version 0.5.0-dev+3455
Commit ce2ca93* (2016-04-08 18:03 UTC)
Platform Info:
  System: Linux (aarch64-unknown-linux-gnu)
  CPU: unknown
  WORD_SIZE: 64
  BLAS: libopenblas (NO_AFFINITY CORTEXA57)
  LAPACK: liblapack
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.0svn-r265783 (ORCJIT, generic)
```

With https://github.com/JuliaLang/julia/pull/15811 and http://reviews.llvm.org/D18508 for other issues.

ginggs:
The remainder of Edmund's patch was for enabling long-double (ld128) variants.

Do we want these?  Are they used by Julia?

yuyichao:
> Do we want these?

Might as well for consistency unless there's reason not to?

ginggs:
@yuyichao 
> Might as well for consistency unless there's reason not to?

Consistency with what?  The list of exported symbols by arch doesn't seem to be consistent.

I did find [this](https://gcc.gnu.org/wiki/Ieee128PowerPC) which seems to be a good reason not to enable **ld128** for powerpc yet.  I will try to find out what I can about armhf and arm64.

ginggs:
I am confused.  It seems that none of builds use **ld128**.  Can someone confirm please?

If this is the case, then this can probably be merged as is.

ViralBShah:
That's right.

PR #140, opened 2016-07-19
Split install into subtargets

arjenroodselaar:
When building a cross-compile toolchain for a freestanding platform one may not
want to build and install dynamic loadables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,8 @@ OLM_MAJOR_SHLIB_EXT := $(SHLIB_EXT).$(SOMAJOR)
 endif
 endif
 
-.PHONY: all check test clean distclean install
+.PHONY: all check test clean distclean \
+	install install-static install-shared install-pkgconfig install-headers
 
 all: libopenlibm.a libopenlibm.$(OLM_MAJOR_MINOR_SHLIB_EXT)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -74,12 +74,19 @@ openlibm.pc: openlibm.pc.in Make.inc Makefile
 	echo "version=${VERSION}" >> openlibm.pc
 	cat openlibm.pc.in >> openlibm.pc
 
-install: all openlibm.pc
+install-static: libopenlibm.a
+	mkdir -p $(DESTDIR)$(shlibdir)
```
tkelman: should probably be libdir?

Makefile
```diff
@@ -74,12 +74,19 @@ openlibm.pc: openlibm.pc.in Make.inc Makefile
...
+	cp -f -a libopenlibm.a $(DESTDIR)$(libdir)/
+
+install-dynamic: libopenlibm.$(OLM_MAJOR_MINOR_SHLIB_EXT) openlibm.pc
```
tkelman: the .pc file isn't strictly for dynamic library use, I don't think

Makefile
```diff
@@ -40,7 +40,7 @@ OLM_MAJOR_SHLIB_EXT := $(SHLIB_EXT).$(SOMAJOR)
 endif
 endif
 
-.PHONY: all check test clean distclean install
+.PHONY: all check test clean distclean install-static install-dynamic install-headers
```
tkelman: I think `install` should also still be in this list

arjenroodselaar:
@tkelman thank you for a review and apologies for the sloppy PR. The comments have been addressed.

Makefile
```diff
@@ -74,12 +75,21 @@ openlibm.pc: openlibm.pc.in Make.inc Makefile
 	echo "version=${VERSION}" >> openlibm.pc
 	cat openlibm.pc.in >> openlibm.pc
 
-install: all openlibm.pc
+install-static: libopenlibm.a
+	mkdir -p $(DESTDIR)$(libdir)
+	cp -f -a libopenlibm.a $(DESTDIR)$(libdir)/
+
+install-shared: libopenlibm.$(OLM_MAJOR_MINOR_SHLIB_EXT) openlibm.pc
```
tkelman: does the openlibm.pc need to be a dependency here?

tkelman:
not sloppy at all, thanks for the contribution!

ViralBShah:
Is this good to merge?

tkelman:
I think openlibm.pc should be deleted from the install-shared line but otherwise yes IMO

arjenroodselaar:
This should be good to go now.

PR #142, opened 2016-08-09
Set optimization level only if not already overridden

staticfloat:
~~Did this edit from a browser, still double-checking this actually works, so don't merge.~~

Tested locally, should be working.


simonbyrne:
Okay, I'll tag a new version. Do I need to bump SOMINOR as this is just a Makefile change?

staticfloat:
No, I wouldn't think so.

PR #145, opened 2016-08-27
Generate libopenlibm.dll.a for windows.

twadleigh:
This is the first step toward addressing https://github.com/JuliaLang/julia/issues/11419.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,12 @@ OBJS =  $(patsubst %.f,%.f.o,\
 	$(patsubst %.S,%.S.o,\
 	$(patsubst %.c,%.c.o,$(filter-out $(addprefix src/,$(DUPLICATE_SRCS)),$(SRCS)))))
 
-# If we're on windows, don't do versioned shared libraries.  If we're on OSX,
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,10 +26,12 @@ OBJS =  $(patsubst %.f,%.f.o,\
 	$(patsubst %.S,%.S.o,\
 	$(patsubst %.c,%.c.o,$(filter-out $(addprefix src/,$(DUPLICATE_SRCS)),$(SRCS)))))
 
-# If we're on windows, don't do versioned shared libraries.  If we're on OSX,
-# put the version number before the .dylib.  Otherwise, put it after.
+# If we're on windows, don't do versioned shared libraries. Also, generate a link library
```
tkelman: isn't this generally called an import library? https://msdn.microsoft.com/en-us/library/windows/desktop/ms682592(v=vs.85).aspx

twadleigh: I'll tweak that.

tkelman:
We should tag this before using it downstream in Julia. I think VERSION will need updating for that.

Issue #5, opened 2012-05-25
olm_ prefixes for openlibm functions


ViralBShah:
@vtjnash, @loladiro, Do you think we should have olm_ prefixes for openlibm functions, so that they don't clash with the system libm?

This could be done through a macro, which allows use of libm names, or with a prefix, depending on what the user wants.

EdSchouten:
The code in https://github.com/JuliaLang/openlibm/pull/77 would actually make this rather easy to support. All source files will include <openlibm.h> and obtain the prototypes through there. We could simply put a macro in that header and use that in all of the sources:

#define OLM_F(name) olm_##

float OLM_F(cosf)(float x) {
  ...
}

It would also make the code slightly cleaner, because we could get rid of the __ieee754_*() garbage at the same time.

ViralBShah:
@EdSchouten This is too much to ask - but if you have tried this, could you please submit a PR? It will be so much cleaner to use code with the prefixes, when one needs to be absolutely sure about calling the right versions.

ViralBShah:
I wondered why the __ieee754_* stuff was there.

ViralBShah:
I am thinking of doing the name mangling, and then linking to the system `-lm` and also simultaneously having the option of using host's fenv.h. That way, we don't have to manage the messy floating point exception stuff, where we are better off using the system provided one.

PR #148, opened 2016-08-31
General cleanups, but also initial bits for 32-bits ARM support

EdSchouten:
For CloudABI's C library, I'm using a somewhat outdated version of openlibm. It would be nice if I could sync in the latest version again. This pull request first does a couple of cleanup changes, so I can sync in an modified copy.

At the same time, I'm working on adding support for using CloudABI on 32-bit ARM systems. 32-bit ARM is interesting: double and long double are the same. The largest supported floating point type is simply 64 bits. This change cleans up the already existent way of doing symbol aliasing to use strong references everywhere. It also adds aliases in places where we forgot to add them.

### DIFF
--- a/bsdsrc/b_tgamma.c
+++ b/bsdsrc/b_tgamma.c
@@ -146,6 +146,9 @@ tgamma(x)
 	else
 		return (neg_gam(x));
 }
+
+OLM_SYMBOL_ALIAS_IF_DOUBLE_IS_LONG_DOUBLE(tgamma, tgammal);
+
 /*
  * Accurate to max(ulp(1/128) absolute, 2^-66 relative) error.
  */
--- a/src/e_acosh.c
+++ b/src/e_acosh.c
@@ -61,3 +61,5 @@ acosh(double x)
 	    return log1p(t+sqrt(2.0*t+t*t));
 	}
 }
+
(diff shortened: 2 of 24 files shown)


nalimilan:
Any idea about the test failures?

EdSchouten:
Hi Milan,

It looks like the build was broken due to two different regressions:

- I forgot to `s/__weak_reference/OLM_SYMBOL_ALIAS/` two source files. This issue wasn't visible on my targets (FreeBSD and CloudABI).

- ARM64: My question is, is this target supposed to work in the first place? Looking at the build output, I can't entirely understand why it would. For example, it tries to use sources from ld80/, which is incorrect. ARM64 uses 128-bit long doubles.

src/math_private.h
```diff
@@ -261,61 +261,6 @@ irint(double x)
 
 #endif /* __GNUCLIKE_ASM */
 
-/*
- * ieee style elementary functions
- *
- * We rename functions here to improve other sources' diffability
- * against fdlibm.
```
tkelman: unfortunate to drop this. probably worth finding a more general solution to https://github.com/JuliaLang/openlibm/issues/5 here

ViralBShah:
I guess this is no longer needed.

PR #150, opened 2016-10-16
Allow setting CC

jackpot51:
This allows CC to be set when calling make, which is useful when cross compiling.

For example, on OS X, you could cross compile for Linux with `ARCH=x86_64 CC=x86_64-elf-gcc make libopenlibm.a`. Without this change, the CC passed to make is overwritten.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -18,6 +18,8 @@ else
   pkgconfigdir = $(libdir)/pkgconfig
 endif
 
+ifndef CC
+
 USEGCC = 1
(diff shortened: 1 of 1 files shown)


jackpot51:
Feel free to close if this does not work correctly or does not meet any needs.

jackpot51:
I have modified it to work the original way unless CROSSCC is set

yuyichao:
> Without this change, the CC passed to make is overwritten.

Does it?

PR #156, opened 2017-03-02
enable -fno-strict-aliasing for clang

simonbyrne:
Type punning via unions in clang is not safe without `-fno-strict-aliasing`. See https://bugs.llvm.org//show_bug.cgi?id=31928.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -39,7 +39,7 @@ AR = ar
 ifeq ($(USECLANG),1)
 USEGCC = 0
 CC = clang
-CFLAGS_add += -fno-builtin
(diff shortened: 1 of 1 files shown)


yuyichao:
= = ..... isn't that the whole point of using `union` in the first place?......................

simonbyrne:
One would think so (gcc does), but apparently the clang authors disagree (and don't bother to document it).

iblis17:
see also: julialang/julia#20454

iblis17:
and ... this http://stackoverflow.com/questions/2958633/gcc-strict-aliasing-and-horror-stories
😱

Issue #157, opened 2017-07-20
aarch64 build is missing frexpl() and related long double functions


mato:
When building openlibm (latest master with #151) on aarch64, the "long double" functions (`frexpl()` and friends) are not built.

I have a (perhaps naive) fix at https://github.com/mato/ocaml-freestanding/commit/9e1d25addb280a2e58f879cf9efacc4e172e56f9.

/cc @staticfloat

staticfloat:
This seems reasonable, @yuyichao do we just avoid calling these long double functions on `aarch64` and that's why we don't have issues with missing symbols?

yuyichao:
We never use these functions anywhere.

`long double` should be [IEEE 754-2008 qual precision on aarch64](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0055b/IHI0055B_aapcs64.pdf) so it should be correct to enable these.

mato:
Shall I open a PR with the change in https://github.com/mato/ocaml-freestanding/commit/9e1d25addb280a2e58f879cf9efacc4e172e56f9?

staticfloat:
Yes please.

mato:
Fixed by #158.

PR #158, opened 2017-07-27
Enable long double functions for aarch64

mato:
Per discussion in #157, it is correct to enable these for aarch64.


mato:
@staticfloat Any chance of getting this merged? I'd like to avoid local changes in my openlibm git subtree for this.

mato:
@staticfloat Thanks!

PR #167, opened 2017-09-21
Allow building on DragonFly BSD

ararslan:
The same change ~~would likely be necessary on OpenBSD as well~~ should _not_ be applied on OpenBSD.

### DIFF
--- a/src/s_nan.c
+++ b/src/s_nan.c
@@ -36,7 +36,7 @@
 
 #include "math_private.h"
 
-#if !defined(__APPLE__) && !defined(__FreeBSD__)
+#if !defined(__APPLE__) && !defined(__FreeBSD__) && !defined(__DragonFly__)
 static __inline int digittoint(int c) {
 	if ('0' <= c && c <= '9')
 		return (c - '0');
--- a/test/Makefile
+++ b/test/Makefile
@@ -3,8 +3,10 @@ include ../Make.inc
 
 # Set rpath of tests to builddir for loading shared library
 OPENLIBM_LIB = -L.. -lopenlibm
-ifneq (,$(findstring $(OS),Linux FreeBSD))
+ifneq ($(OS),WINNT)
+ifneq ($(OS),Darwin)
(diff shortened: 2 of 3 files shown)


test/Makefile
```diff
@@ -3,7 +3,7 @@ include ../Make.inc
 
 # Set rpath of tests to builddir for loading shared library
 OPENLIBM_LIB = -L.. -lopenlibm
-ifneq (,$(findstring $(OS),Linux FreeBSD))
+ifneq (,$(findstring $(OS),Linux FreeBSD DragonFly))
```
ararslan: Actually, this should probably just be
```make
ifneq ($(OS),Windows_NT)
ifneq ($(OS),Darwin)
OPENLIBM_LIB += -Wl,-rpath=$(OPENLIBM_HOME)
endif
endif
```
That would likely cover the bases for any other BSD we want to support.

src/s_nan.c
```diff
@@ -36,7 +36,7 @@
 
 #include "math_private.h"
 
-#if !defined(__APPLE__) && !defined(__FreeBSD__)
+#if !defined(__APPLE__) && !defined(__FreeBSD__) && !defined(__DragonFly__)
```
ararslan: Should this maybe just be a check on Windows and Linux? Presumably all BSDs should be excluded here.

test/Makefile
nalimilan: Why not, better use the case which is most likely to work for unlisted platforms as a fallback.

src/s_nan.c
nalimilan: Isn't there a common flag to detect any kind of BSD? Here the test is clearly specific to a family of OSes.

ararslan: Interestingly, this builds fine on OpenBSD without changing this line, and adding `!defined(__OpenBSD__)` actually causes the tests to fail with an undefined reference to `digittoint`. So I guess this is better left as an explicit list.

PR #170, opened 2017-11-15
Make exp2() work with icc

certik:
Make the code more readable.

I provided two versions, both work with icc. One is very fast and equivalent to the original, but it is not very readable. The other is a simple mathematical form, however compilers might generate a bit slower code than the original. So for now I provided both options, and one can select at compile time.

### DIFF
--- a/src/s_exp2.c
+++ b/src/s_exp2.c
@@ -306,6 +306,31 @@ static const double tbl[TBLSIZE * 2] = {
 	0x1.690f4b19e9471p+0,	-0x1.9780p-45,
 };
 
+
+/*
+ * Compute round(x*2^p)/2^p
+ *
+ * Contains two versions that one can select at compile time. One is a
+ * platform independent mathematical form, that might be perhaps abit slower.
+ * The other is a very fast version, that depends on the double precision
+ * floating point representation. Both versions compile with `-ffast-math` or
+ * an equivalent compiler option.
+ */
+double ldexp_floor_ldexp(double x, int p)
+{
+#if 0
+    // Mathematical form
+	return ldexp(floor(ldexp(x, p)+0.5), -p);
+#else
+    // Fast form
+    double r = 0x1.8p52 / (1 << p);
+    volatile double t;
+    t = x + r;
+    return t - r;
+#endif
+}
+
+
 /*
  * exp2(x): compute the base 2 exponential of x
  *
@@ -369,7 +394,7 @@ exp2(double x)
 	i0 += TBLSIZE / 2;
 	k = (i0 >> TBLBITS) << 20;
 	i0 = (i0 & (TBLSIZE - 1)) << 1;
-	t -= redux;
+	t = ldexp_floor_ldexp(x, TBLBITS);
 	z = x - t;
 
 	/* Compute r = exp2(y) = exp2t[i0] * p(z - eps[i]). */


src/s_exp2.c
```diff
@@ -306,6 +306,31 @@ static const double tbl[TBLSIZE * 2] = {
...
+ */
+double ldexp_floor_ldexp(double x, int p)
+{
+#if 0
+    // Mathematical form
+	return ldexp(floor(ldexp(x, p)+0.5), -p);
+#else
+    // Fast form
+    double r = 0x1.8p52 / (1 << p);
+    volatile double t;
```
yuyichao: This requires spilling and this function isn't even static I highly doubt this is fast everywhere.

certik:
That's true. I'll have to put the "volatile" into a macro, and only
enable for icc.

Do you have any tips how to write this without volatile for icc? The
only other idea that I have is to write that in assembly.

yuyichao:
> Do you have any tips how to write this without volatile for icc? The only other idea that I have is to write that in assembly.

No. I don't use icc and I don't think case by case workaround like this is maintainable at all.

yuyichao:
Or in another word, if there's a way to write the code so that it makes all supported compiler on all supported archs generate at least the same code we can certainly merge it (although even in that case we will will assume ieee semantic here and won't guarantee the workaround to work in future versions.) If it requires compiler version check just to workaround the icc issue, we should just enable the ieee conforming mode for icc instead.

vtjnash:
I recommend compiling with `icc -fp-model precise` to enable standards-compliance mode. As Yichao mentioned, a volatile store will likely incur a significant performance penalty (especially since this is spilling an fp-reg, and not just an ireg, so it has much higher latency and lower throughput per http://agner.org/optimize/instruction_tables.pdf).

certik:
As I said, I agree, I didn't realize it was spilling a register. I apologize.

@vtjnash yes, I have already verified that `icc` works if you use `-fp-model=precise`. All tests pass.

Given that you will assume IEEE semantics, and as @yuyichao said, you will not guarantee this to work in future versions, I think the best is to just keep the original code, since that already follows IEEE semantics anyway. So I am going to close the PR.

My work here was to fix the code to not rely on IEEE semantics and to show that it is possible, without sacrificing speed ---- and yes, as I said, I didn't realize that volatile spills the register, so one would need to provide an assembly implementation of this routine, similar to, e.g., this one:

https://github.com/JuliaLang/openlibm/blob/b2520473c2b2f755ddf385f366fee097f68b71ba/amd64/s_scalbn.S#L30

Unless there is some other way to write this in C.

Regarding maintainability, that is up to you to decide, and looks like @yuyichao made clear above that it is not maintainable. So that settles it.

For my own projects, I don't rely on IEEE semantics, and it's easy to make it maintainable --- for example in this case we need a fast "intrinsic" for doing the `ldexp(floor(ldexp(x, p)+0.5), -p)` operation. I typically leave it in this form, but if speed is really a concern, then I would provide a fast implementation of this, and call it `ldexp_floor_ldexp(x, p)`, exactly as I have done in this PR, except that, as you pointed out, my "fast" implementation isn't quite as fast, so I would try to figure out a better way to write this, either in C, or, if there is no other way, in assembly. But since this is not the direction you guys want to go, that is perfectly fine with me. This is more of an academic exercise for me, as I was trying to show to people in another thread that it is possible to fix a real project, with a real problem that `icc` miscompiles, and to fix it in a way so that `icc` works by default on it. And this PR shows exactly how to do that, and this comment shows how one can take this further to make it as fast as original, if that was the task at hand.

Thank you guys for the review!

yuyichao:
(OT)
> https://github.com/JuliaLang/openlibm/blob/b2520473c2b2f755ddf385f366fee097f68b71ba/amd64/s_scalbn.S#L30

eriously?

yuyichao:
(Had control hold down while changing a line with Enter just now......................)

OT

> https://github.com/JuliaLang/openlibm/blob/b2520473c2b2f755ddf385f366fee097f68b71ba/amd64/s_scalbn.S#L30

Seriously? We have assembly on x64 using x87 instructions? We should almost certainly get rid of those...............

> For my own projects, I don't rely on IEEE semantics

I do too, in many cases. And i do think that's a very valid need. I recently got a 30% speedup with fastmath on sincos computation with a 2e-7 max error in an application that I care no more than 1e-4 (which, if you think about it, is not a good trade off in general).

I do agree what you want to do is vaid if the project is mainly about high level computation though a much better way to do it is to annotate when you want or not want IEEE semantics. Both GCC and julia (not sure about clang/icc) provide ways to do this on a function local/scope level (and julia's support for this will be greately improved on llvm 5.0). However, if the whole point is to compute things accurately, like what is done here as well as a lot of other scientific computations, the default should be ieee complaint.

certik:
@yuyichao yes, exactly. I just finished writing a long comment here:

https://github.com/JuliaLang/julia/issues/24568#issuecomment-344643007

I am very proud it, as I think I figured out that there are two schools, an IEEE school and a Fortran school, and essentially in this PR I was trying to convert openlibm from IEEE school to Fortran school, while you wanted to keep it in IEEE school. My comment there explains exactly the details of each school. 

openlibm *could* be converted to Fortran school --- the way you would do it is along this PR, simply lift the code to not rely on IEEE, and then provide fast intrinsics/building blocks (that rely on IEEE) that you can use. Ultimately, `exp2` and other functions just have some numerical algorithm to compute them, and that can definitely be done in Fortran school. But openlibm can also stay fully in the IEEE school, in fact it probably makes more sense, because you have to handle things like NaN, infinities etc. and those are only well defined in the IEEE school. So you might as well simply assume IEEE everywhere.

This is exactly what you said, I just rewrote your argument using my words. As I am trying to figure this out myself.

yuyichao:
> openlibm could be converted to Fortran school --- the way you would do it is along this PR, simply lift the code to not rely on IEEE, and then provide fast intrinsics/building blocks (that rely on IEEE) that you can use. 

Well, "could" is a really vague term. There's no "simply" lifting the code here and this whole library is exactly that building block.

certik:
@yuyichao yes, you can't lift the NaN and infinities, as I said. (You can, however, lift some intermediate calculations, as I have shown in this PR, just not the whole library.) So you might as well simply assume IEEE everywhere and treat the whole library as that building block, as you said.

simonbyrne:
This code absolutely assumes IEEE semantics, anything else is way too dangerous for a math library. I think the way forward would be adding a special CFLAG to the makefile for icc.

PR #171, opened 2017-11-16
WIP: How to speedup exp2

certik:
This is an example pull request that shows how to speedup `exp2` using `-ffast-math` in a portable way. I don't intend this to be merged, it's just to show how to do it.
```
Original optimization level:
0.588
Original optimization + round(x*256)/256.:
0.923
Original optimization + round256(x):
0.584
-ffast-math + round(x*256)/256.:
0.667
-ffast-math + round256(x):
0.431
-ffast-math + round256(x) + no inf/nan checks:
0.375
```
I split the IEEE dependent part into a separate module and use link time optimization so that it gets inlined. The IEEE dependent part is compiled without `-ffast-math`, the rest is compiled with `-ffast-math`. The `round256(x)` does the equivalent thing as before.

The `-ffast-math` delivers significant speedup over the original optimization, as expected. The code is essentially equivalent.

For comparison, I also benchmarked using the slower but IEEE independent `round(x*256)/256.`, and with `-ffast-math`, the performance is actually not bad, but not as good as the optimized `round256()`.

Finally, I also removed all the inf/nan checks using the following patch:
```diff
--- a/src/s_exp2.c
+++ b/src/s_exp2.c
@@ -347,25 +347,6 @@ exp2(double x)
        u_int32_t hx, ix, lx, i0;
        int k;

-       /* Filter out exceptional cases. */
-       GET_HIGH_WORD(hx,x);
-       ix = hx & 0x7fffffff;           /* high word of |x| */
-       if(ix >= 0x40900000) {                  /* |x| >= 1024 */
-               if(ix >= 0x7ff00000) {
-                       GET_LOW_WORD(lx,x);
-                       if(((ix & 0xfffff) | lx) != 0 || (hx & 0x80000000) == 0)
-                               return (x + x); /* x is NaN or +Inf */
-                       else 
-                               return (0.0);   /* x is -Inf */
-               }
-               if(x >= 0x1.0p10)
-                       return (huge * huge); /* overflow */
-               if(x <= -0x1.0ccp10)
-                       return (twom1000 * twom1000); /* underflow */
-       } else if (ix < 0x3c900000) {           /* |x| < 0x1p-54 */
-               return (1.0 + x);
-       }
-
        /* Reduce x, computing z, i0, and k. */
        STRICT_ASSIGN(double, t, x + redux);
        GET_LOW_WORD(i0, t);
```
And as you can see, now we are starting to pick up some speed. The reason the benchmark is so sensitive to simple changes like that is that the function is already *very* fast.

Anyway, as you can see, I haven't done anything, just gave up IEEE compliance, but I haven't even modified any code, just moved the IEEE dependent part to a separate module, and got rid of infinities/nan checks, and now the original is more than 1.5x slower. That is a pretty heavy performance price to pay for IEEE compliance. @StefanKarpinski, you said several times that it is possible to match performance of `-ffast-math` while keeping IEEE compliance. I don't think that's true, because by getting rid of the inf/nan checks, I got significant speedup. You said there is no price to pay, but there is. But even without it, if you can match it, I would be interested in how.

I used gcc + gfortran 6.3.0 in Ubuntu 17.04 and it was ran on my laptop, the processor is Intel(R) Xeon(R) CPU E3-1505M v6 @ 3.00GHz. You are welcome to check my work, hopefully I didn't make any mistakes.

### DIFF
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -4,6 +4,7 @@ set -e
 set -x
 
 FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
+FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -O3 -march=native -ffast-math -funroll-loops"
 
 gcc -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
 gfortran $FFLAGS -c test_exp2.f90 -o test_exp2.o
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -3,6 +3,10 @@
 set -e
 set -x
 
-cd ..
+FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
 
-gcc -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -Iinclude -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c src/s_exp2.c -o src/s_exp2.c.o
+gcc -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
+gfortran $FFLAGS -c test_exp2.f90 -o test_exp2.o
+gfortran $FFLAGS s_exp2.c.o test_exp2.o -o test_exp2
+
+./test_exp2
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -7,12 +7,15 @@ FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
 FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -O3 -march=native -ffast-math -funroll-loops"
 
 CFLAGS="-fno-gnu89-inline -fno-builtin -fPIC -m64 -std=c99 -Wall -O3"
-gcc -flto $CFLAGS -Wno-implicit-function-declaration -c redux.c -o redux.o
+CFLAGS="$CFLAGS -march=native -funroll-loops"
+#LTO="-flto"
+gcc $LTO $CFLAGS -Wno-implicit-function-declaration -c redux.c -o redux.o
 
-#CFLAGS="$CFLAGS -march=native -ffast-math -funroll-loops"
+#CFLAGS="$CFLAGS -ffast-math"
 
-gcc -flto $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
-gfortran -flto $FFLAGS -c test_exp2.f90 -o test_exp2.o
-gfortran -flto $FFLAGS s_exp2.c.o test_exp2.o redux.o -o test_exp2
+gcc $LTO $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
+gcc $LTO $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/e_pow.c -o e_pow.c.o
+gfortran $LTO $FFLAGS -c test_exp2.f90 -o test_exp2.o
+gfortran $LTO $FFLAGS test_exp2.o s_exp2.c.o e_pow.c.o redux.o -o test_exp2
 
 ./test_exp2
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -6,7 +6,10 @@ set -x
 FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
 FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -O3 -march=native -ffast-math -funroll-loops"
 
-gcc -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
+CFLAGS="-fno-gnu89-inline -fno-builtin -fPIC -m64 -std=c99 -Wall -O3"
+#CFLAGS="$CFLAGS -march=native -ffast-math -funroll-loops"
+
+gcc $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
 gfortran $FFLAGS -c test_exp2.f90 -o test_exp2.o
 gfortran $FFLAGS s_exp2.c.o test_exp2.o -o test_exp2
 
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -7,10 +7,12 @@ FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
 FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -O3 -march=native -ffast-math -funroll-loops"
 
 CFLAGS="-fno-gnu89-inline -fno-builtin -fPIC -m64 -std=c99 -Wall -O3"
+gcc -flto $CFLAGS -Wno-implicit-function-declaration -c redux.c -o redux.o
+
 #CFLAGS="$CFLAGS -march=native -ffast-math -funroll-loops"
 
-gcc $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
-gfortran $FFLAGS -c test_exp2.f90 -o test_exp2.o
-gfortran $FFLAGS s_exp2.c.o test_exp2.o -o test_exp2
+gcc -flto $CFLAGS -I../include -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c ../src/s_exp2.c -o s_exp2.c.o
+gfortran -flto $FFLAGS -c test_exp2.f90 -o test_exp2.o
+gfortran -flto $FFLAGS s_exp2.c.o test_exp2.o redux.o -o test_exp2
 
 ./test_exp2
--- a/bench/build.sh
+++ b/bench/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -e
+set -x
+
+cd ..
+
+gcc -fno-gnu89-inline -fno-builtin -O3 -fPIC -m64 -std=c99 -Wall -Iinclude -DASSEMBLER -D__BSD_VISIBLE -Wno-implicit-function-declaration -c src/s_exp2.c -o src/s_exp2.c.o
--- a/bench/redux.c
+++ b/bench/redux.c
@@ -1,0 +1,5 @@
+double round256(double x)
+{
+    double redux = 0x1.8p52 / 256;
+    return (x + redux) - redux;
+}
--- a/bench/test_exp2.f90
+++ b/bench/test_exp2.f90
@@ -1,0 +1,18 @@
+program test_exp2
+use iso_c_binding, only: c_double
+implicit none
+integer, parameter :: dp = kind(0.d0)
+
+interface
+    real(c_double) function c_exp2(x) bind(c, name="exp2")
+    import :: c_double
+    real(c_double), value, intent(in) :: x
+    end function
+end interface
+
+real(dp) :: y
+
+y = c_exp2(0.7_dp)
+print *, y
+
+end program
--- a/bench/test_exp2.f90
+++ b/bench/test_exp2.f90
@@ -15,7 +15,7 @@ end interface
 real(dp), allocatable :: x(:), r(:), r_exact(:), err_abs(:)
 integer, allocatable :: err_ulp(:)
 real(dp) :: t1, t2, x_min, x_max
-integer :: i, n
+integer :: i, j, n, n2
 integer, allocatable :: seed(:)
 
 call random_seed(size=n)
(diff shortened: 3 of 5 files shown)


certik:
I should mention, that by turning on `-ffast-math`, the checks change from
```
 Max abs error:   3.6304123742133376E+280
 Max ulp error:           1 ulp
 Histogram
           0       99963   99.963000000000008      %
           1          37   3.6999999999999998E-002 %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
to
```
 Max abs error:                  Infinity
 Max ulp error:           1 ulp
 Histogram
           0       99952   99.951999999999998      %
           1          26   2.6000000000000002E-002 %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
In other words, on 100,000 random values from the original range (-1075, 1024) where the `exp2` function returns non zero and non infinity, 99.963% of cases gives the correct answer with 0 ulp error, and there are 37 cases where the error is 1 ulp. With `-ffast-math`, the percentages don't really change, but  there are 11 less values that got 0 ulp, 11 less values that got 1 ulp, and I think 11+11=22 values that return an infinity. These numbers do not change when doing the options above like round256 or round, or with or without inf/nan checks. They only change when doing with or without `-ffast-math`.

To fix it, I reduced the maximum allowable value from 1024 to 1023 with this patch:
```diff
--- a/bench/test_exp2.f90
+++ b/bench/test_exp2.f90
@@ -25,7 +25,7 @@ allocate(x(n), r(n), r_exact(n), err_abs(n), err_ulp(n))

 call random_number(x)
 x_min = -1075
-x_max = 1024
+x_max = 1023
 x = x*(x_max - x_min) + x_min
 print *, "Running c_exp2()"
 call cpu_time(t1)
```
now the original routine gives:
```
 Max abs error:   6.7621699985365152E+271
 Max ulp error:           1 ulp
 Histogram
           0       99954   99.954000000000008      %
           1          46   4.5999999999999999E-002 %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
By switching to `round256` but keeping the original optimization I get no change:
```
 Max abs error:   6.7621699985365152E+271
 Max ulp error:           1 ulp
 Histogram
           0       99954   99.954000000000008      %
           1          46   4.5999999999999999E-002 %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
and with `-ffast-math` and `round256` I get:
```
 Max abs error:   3.2244539253885818E+265
 Max ulp error:           1 ulp
 Histogram
           0       99975   99.975000000000009      %
           1          25   2.5000000000000001E-002 %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
So `-ffast-math` has *improved* accuracy, but it can just be a statistical fluke. Removing the inf/nan checks now doesn't change anything.

So the only downside of `-ffast-math` is that for 1023 < x < 1024, it gets to infinity a bit sooner. Not relevant in practice.

simonbyrne:
Thanks, that was interesting. Note that in general testing using the same precision isn't terribly useful for checking such fine-grained accuracy: 0 ulps here is slightly misleading since it is being compared to a value that has already been rounded. In reality you need a higher-precision library (e.g. quadmath or MPFR) as a reference.

I'm not too worried in this case though as the histogram shows a drop in 1 ulp values. My guess that this is due to wider use of FMA operations, which both increase accuracy and speed (though aren't available on all processors). Unfortunately, as we explained in the other thread, C doesn't have a great way to express this other than via compile time flags such as `fp-contract` or `fastmath`. This is part of the reason we started implementing them ourselves in Julia, e.g. pure-Julia `exp` [sees a similar speedup of 1.3-1.5x](https://github.com/JuliaLang/julia/pull/19831).

Unfortunately, as @yuyichao explained on the other thread, there is no way we can enable `fastmath` generally: it is simply too dangerous, as you lose your ability to reason about accuracy and establish formal results.  It would be interesting to see if you get a similar speedup with just `fp-contract` (which is slightly safe, though I'm still hesitant to enable it more widely as it can occasionally give strange results).

certik:
@simonbyrne thank you. Note that I use quadruple precision to check this. Then I round it to double precision. So the results should be reliable. The other way to do it is not to round the quadruple precision number `r_exact` and compute the difference in quadruple precision. Then you can get fractional ulp. But I figured that we are not interested how accurate it is compared to the exact answer, but rather how far the exp2 is from the exact answer that is correctly rounded to double precision. So that's what I did.

I'll try the `fp-contract`.

simonbyrne:
> But I figured that we are not interested how accurate it is compared to the exact answer, but rather how far the exp2 is from the exact answer that is correctly rounded to double precision.

Actually, I think the former is more informative, as it tells you e.g.:
- exact error <0.5 ulps =>  correctly rounded.
- exact error <1ulp => one of the 2 nearest floats.

certik:
@simonbyrne so `-ffp-contract=fast` is the default in gcc. I did enable it by hand just in case, but observed no difference in timings for the original code.

bench/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
+set -x
+
+FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
+FFLAGS="-Wall -Wextra -Wimplicit-interface -fPIC -O3 -march=native -ffast-math -funroll-loops"
```
vtjnash: adding `-march=native` isn't a fair comparison

certik: @vtjnash I think you put your comment on the wrong line. This is a Fortran flag, which is irrelevant for openlibm. But your comment applies to the CFLAGS below. That's a good point, I'll rerun the benchmarks to see.

certik: @vtjnash is `-march=native` "allowed" for openlibm?

vtjnash: Sure. It's just a lot more work to run the benchmarks that way (and should be effectively a constant factor, similar to how switching processors would alter the benchmarks).

yuyichao: > is `-march=native` "allowed" for openlibm?

For any project that's not targeted for a very specific use this should not be an option specified by default. It is perfectly fine as a user supplied option. It's acceptable when doing benchmarks if you make sure everything else are doing the same.

simonbyrne:
If you're using gcc, then `-ffast-math` is equivalent to:
```
-fno-trapping-math -funsafe-math-optimizations -ffinite-math-only -fno-errno-math -fno-signaling-nans -fno-rounding-math -fcx-limited-range -fno-signed-zeros
```
It would be useful to know which of those is responsible for the speedup, as that might give us some insight into what it is doing.

bench/build.sh
certik: Ok. Just to clarify what I did --- I took exactly the default options that openlibm itself provides. As I assume those are the ones that you want people to use when doing benchmarking. So I will now add `-march=native` to those as well. If there is any other option that you think I should add to make a fair comparison for the best openlibm IEEE speed, let me know.

certik:
I tried it on another machine, and I got:

CPU: Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz
gcc 7.2.0

```
original
1.2639999999999998
original + round256
1.3000000000000003
-ffast-math + round256
1.1960000000000002
-ffast-math + round256 + no inf/nan
1.1639999999999997
```
These benchmarks are too close to each other to be able to infer anything.


Changed the benchmark loop structure to benchmark a single element many
times, and just do a short array. This eliminates the slower memory speed
problem on this computer.
```
original
7.2480000000000002
original + round256
7.2160000000000002
-ffast-math + round256
4.8639999999999999
-ffast-math + round256 + no inf/nan
1.1719999999999999
```
This is too much of a speedup simply be removing an if statement. As such, I think I might have fooled myself here and also above. I am suspecting that the `-flto` inlined the `exp2` function into my benchmark code, but only sometimes, based on the optimizations options that I give it. Consider this:
```
-fno-math-errno -funsafe-math-optimizations -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fcx-limited-range 
1.2360000000000000
-fno-rounding-math -fno-signaling-nans -fcx-limited-range
6.7199999999999998

-fno-math-errno -funsafe-math-optimizations -ffinite-math-only
6.7439999999999998

-fno-math-errno -funsafe-math-optimizations -ffinite-math-only -fno-rounding-math -fno-signaling-nans
6.6680000000000001
```

Essentially the speed drops by removing *any* of the `-ffast-math` options. I think the speed drop is due to `-flto` deciding not to inline `exp2`.

So I dropped `-flto`, and redid benchmarks:
```
original no inf/nan
8.7200000000000006
original no inf/nan round256
9.7080000000000002
-ffast-math no inf/nan round256
10.956000000000000
```
Now the performance drops when using my `round256` function as expected, as it doesn't get inlined. But the performance should not drop when turning on the `-ffast-math` option, but it does. 
```

So I switched to Intel. And roughly the same behavior. The `-fp-model=strict` is faster than the default `icc` options.

I did notice even before though, that just changing the `exp2` function to:
```c
OLM_DLLEXPORT double
exp2(double x)
{
    return x;
}
```
takes about 25% of the time. So the function is so fast that it's only about 4x slower than doing nothing.  So if the function is doing 4x nothing, then that might not be a good example for running these kinds of benchmarks, as one then has to be extremely careful how to benchmark it, not to fool myself like I think I did.

I did remove the inf/nan checks for these tests.


So right now I think the `-ffast-math` is not helping at all on `exp2`.

certik:
P.S. the `-march=native` speeds `exp2` a bit, but not as much as the other changes.

certik:
The `exp2` results are inconclusive --- a much more careful benchmarking job must be done on such a short function. I switched to the `__ieee754_pow` function defined in `e_pow.c`, which is a longer function, and also works out of the box with `-ffast-math`. First let's check accuracy. Everything below is doing pow(2, x), which is equivalent to exp2(x), so I can keep my infrastructure. I didn't see any special casing for 2 in `pow`, so this should be representable.

Original:
```
 Max abs error:   1.9958403095347198E+292
 Max ulp error:           1 ulp
 Histogram
           0     9075053   90.750529999999998      %
           1      924947   9.2494699999999987      %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
and `-ffast-math`:
```
 Max abs error:   1.9958403095347198E+292
 Max ulp error:           1 ulp
 Histogram
           0     7477687   74.776869999999988      %
           1     2522313   25.223129999999998      %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
So about 15% of the exact 0ulp results are now 1ulp. That is very acceptable for me, since even before about 10% of all results were 1ulp, so now it's 25%. 

Let's look at the benchmarks. No LTO.
 ```
empty routine
4.4000000000000011E-002
original
1.6279999999999999
-ffast-math
1.5759999999999998
```
I ran each many times and took the best timing. There is only very slight speedup with `-ffast-math` if any at all. The numbers fluctuate more than the difference. So I don't think it's significant.

File: bench/build.sh (hunk 1 above)
Comment (certik) [reply bench/build.sh]: Anyway, I moved the `-march=native` to the CFLAGS, it provides a bit of a speedup, but not much.

**Comment by certik** [comment]
I tried Intel as well, I don't see any statistically significant speedup of using the default options vs `-fp-model=strict`. 

It looks the openlibm math subroutines don't benefit at all from `-ffast-math`. They are already written very efficiently.

@simonbyrne I still wonder if I got some speedup on my other before computer, since on my current computer I don't have the fma instruction I think. But my previous computer has it I think.

I am closing this issue. For openlibm I don't think it's worth it to use `-ffast-math`.

**Comment by simonbyrne** [comment]
That makes sense, thanks for trying it out.

**Comment by certik** [comment]
@simonbyrne hm, I thought this is settled, but just in case I tried this one more time on my laptop at home. I did the `pow` function, since that does not require any LTO, and works out of the box with `-ffast-math`. And I got significant speedup!!

With default options + `-march=native -funroll-loops` I get:
0.66800000000000004  (90% is 0ulp, 10% is 1ulp)
With `-ffast-math` I get:
0.49199999999999999 (75% is 0ulp, 25% is 1ulp)

I tested this several times, it seems reproducible. I traced it to the `-funsafe-math-optimizations` option, everything else can be ignored. But my man page says that `-funsafe-math-optimizations` is equal to `-fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math`, but not so when I test it:

-funsafe-math-optimizations
0.49199999999999999
-fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math
0.61199999999999999

Any ideas how to figure out exactly which option is responsible for the speedup?

**Comment by certik** [comment]
I verified in the assembly that the -funsafe-math-optimizations option adds vfmadd213sd, which is fma. So that's the reason for the speedup. There are some other rearrangements in the assembly, those are probably unrelated.

**Comment by certik** [comment]
I figured it out. So these are the minimal options that are needed:
```
-ffp-contract=fast -fno-signed-zeros -fno-trapping-math -fassociative-math
0.49199999999999999
```
if I take any of them out, I will lose performance.

**Comment by certik** [comment]
As an example, if I only do `-ffp-contract=fast`, then I verified in the assembly that fma instructions get added, but the overall performance is somewhere in the middle:
0.59199999999999997
So the other options are still needed.

**Comment by certik** [comment]
Ok, here is a tough one for you @yuyichao: if I only enable `-ffp-contract=fast`, then the accuracy changes from:
```
 Time:  0.66800000000000004
 Computing Errors
 Done
 Max abs error:   1.9958403095347198E+292
 Max ulp error:           1 ulp
 Histogram
           0     9072633   90.726329999999990      %
           1      927367   9.2736699999999992      %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
to
```
 Time:  0.58399999999999996
 Computing Errors
 Done
 Max abs error:   1.9958403095347198E+292
 Max ulp error:           1 ulp
 Histogram
           0     9072627   90.726269999999985      %
           1      927373   9.2737299999999987      %
           2           0   0.0000000000000000      %
           3           0   0.0000000000000000      %
           4           0   0.0000000000000000      %
```
and the code got faster. It's about half of the speedup, the other half is achieved by the other "non-safe" options above, but in this comment let's not discuss those.

Of the 10,000,000 values, 6 moved from 0ulp to 1 ulp.  Obviously, I don't care about such a small change at all, but you do. The maximum error bound might have changed. Perhaps some values from 1ulp moved to 2ulp, who knows? I didn't find any with my test, but that does not mean they are not there. It's a matter of principle. 

Moreover, these results on my laptop are now different than those on my desktop. So openlibm now lost the IEEE guarantee of the same results on all platforms.  So you can't use fma instructions. 

But if you don't use fma instructions, then you are leaving performance on the table. 

How do you resolve this? Will you sacrifice this speedup for the IEEE guarantee, or will you sacrifice IEEE guarantee across fma and non-fma CPUs to gain this performance gain?

yuyichao:
Not sure why it's a "tough one" for me since it is exactly covered by what I said before.

Any reasonable and precisely specifiable error bound are fine. This includes value changing compiler options are certainly fine as long as it's not fast-math and you can prove that the value changes are still within a bound. Options that changes evaluation without requiring isa extension should still preferably be applied manually rather than letting the compiler do it. Options that benefit from isa extension is certainly a case compiler options can be used to workaround language limitations (and gcc support doing this in a local scope) and we are already doing so in the julia version.

certik:
So you will get different results on my laptop (with fma) and my desktop (without fma)? I thought one of the advantages of using an IEEE compiler is that you get exactly the same numbers on all platforms.

PR #175, opened 2018-01-29
Fix reversed outputs of sincosl

takuyanakaoka:
Hi,

I found that the outputs from sincosl are reversed.
According to man(3), the second argument is sin and the third argument is cos.

Thanks.


simonbyrne:
Yup, looks like a type: I guess the variable names give it away.

PR #176, opened 2018-04-05
Fixes for building on OpenBSD

ararslan:


Makefile
```diff
@@ -46,6 +46,13 @@ OLM_MAJOR_SHLIB_EXT := $(SHLIB_EXT).$(SOMAJOR)
 endif
 endif
 
+# OpenBSD doesn't support cp -a, so we'll use the equivalent
+ifeq ($(OS),OpenBSD)
+CP_A := cp -RpP
+else
+CP_A := cp -a
+endif
```
staticfloat: I actually think `cp -RpP` will work everywhere; so we can probably just use that.

ararslan:
I've separated this out into three distinct commits:

1. Use Clang on OpenBSD
2. Use `cp -RpP` everywhere, as suggested
3. Hide a bunch of stuff behind `#ifndef`s to avoid Clang warnings on OpenBSD

PR #177, opened 2018-04-06
Add Windows testing with AppVeyor

ararslan:
This runs on my AppVeyor account to avoid using up the JuliaLang paid AV queue.

Closes #165.

PR #178, opened 2018-04-07
msun: signed overflow in atan2

ararslan:
Incorporated from FreeBSD source tree: https://github.com/freebsd/freebsd/commit/b21ccf63f28a3a4692d8a31419e0a725a1b1a800. The original authorship has been preserved here. Commit message from upstream:

> As a component of atan2(y, x), the case of x == 1.0 is farmed out to
> atan(y). The current implementation of this comparison is vulnerable
> to signed integer underflow (that is, undefined behavior), and it's
> performed in a somewhat more complicated way than it need be. Change
> it to not be quite so cute, rather directly comparing the high/low
> bits of x to the specific IEEE-754 bit pattern that encodes 1.0.
>
> Note that while there are three different e_atan* files in the
> relevant directory, only this one needs fixing. e_atan2f.c already
> compares against the full bit pattern encoding 1.0f, while
> e_atan2l.cuses bitwise-ands/ors/nots and so doesn't require a change.

### DIFF
--- a/src/e_atan2.c
+++ b/src/e_atan2.c
@@ -71,7 +71,7 @@ __ieee754_atan2(double y, double x)
 	if(((ix|((lx|-lx)>>31))>0x7ff00000)||
 	   ((iy|((ly|-ly)>>31))>0x7ff00000))	/* x or y is NaN */
 	   return x+y;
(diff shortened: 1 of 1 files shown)


ararslan:
The Windows failures are because #177 is still incomplete.

PR #130, opened 2016-04-07
Enable optimization again for *int.c and *intf.c 

ginggs:
Optimization was disabled in dcc0d8da7400b68cb0b02c4605ce10fd605f3a1a.  This PR serves as a reminder to enable it again.

PR #180, opened 2018-05-26
Bump version to 0.5.6

ginggs:


ginggs:
Can we get #130 in?

ginggs:
#181 looks good, too.

PR #190, opened 2019-01-14
Fix compilation of gcc when using openlibm as system libm

jackpot51:
For some sad reason, gcc poisons LONG_DOUBLE in sreal.c. The identifier has been changed to OPENLIBM_LONG_DOUBLE

### DIFF
--- a/include/openlibm_math.h
+++ b/include/openlibm_math.h
@@ -26,7 +26,7 @@
 #endif
 
 #ifndef __arm__
-#define LONG_DOUBLE
+#define OPENLIBM_LONG_DOUBLE
 #endif
 
 #ifndef __pure2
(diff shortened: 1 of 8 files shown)


include/openlibm_math.h
```diff
@@ -26,7 +26,7 @@
 #endif
 
 #ifndef __arm__
-#define LONG_DOUBLE
+#define OPENLIBM_LONG_DOUBLE
```
ararslan: Can we make this `OLM_LONG_DOUBLE`? We use the `OLM_` prefix in other places so it would be nice to be consistent.

jackpot51: Thanks, I will do that

jackpot51:
@ararslan it is ready!

PR #192, opened 2019-04-13
wasm32 support

minad:
Build openlibm with clang-8 for wasm32. There is no libc dependency. Functions depending on fenv are not provided.


ViralBShah:
Looks good

PR #197, opened 2019-10-25
NetBSD fix

gufe44:
https://github.com/JuliaMath/openlibm/issues/196

### DIFF
--- a/src/cdefs-compat.h
+++ b/src/cdefs-compat.h
@@ -14,7 +14,7 @@
 #ifdef __GNUC__
 #ifndef __strong_reference
 #ifdef __APPLE__
-#define __strong_reference(sym,aliassym) __weak_reference(sym,aliassym)
+#define __strong_reference(sym,aliassym) __weak_reference_compat(sym,aliassym)
 #else
 #define __strong_reference(sym,aliassym)	\
 	OLM_DLLEXPORT extern __typeof (sym) aliassym __attribute__ ((__alias__ (#sym)));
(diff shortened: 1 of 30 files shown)


ViralBShah:
@keno ?

staticfloat:
LGTM

src/cdefs-compat.h
```diff
@@ -14,21 +14,27 @@
...
 
 #ifdef __wasm__
-#  define __weak_reference(sym,alias) __strong_reference(sym,alias)
+#  define __weak_reference_compat(sym,alias) __strong_reference(sym,alias)
 #endif
 
-#ifndef __weak_reference
+#ifdef __weak_reference
+#ifdef __NetBSD__
+#define __weak_reference_compat(sym,alias) __weak_reference(sym)
```
vtjnash:
```suggestion
#define __weak_reference_compat(sym,alias) __weak_alias(alias,sym)
```

src/cdefs-compat.h
```diff
@@ -50,17 +56,17 @@
 #endif	/* __STDC__ */
 #elif defined(__clang__) /* CLANG */
 #ifdef __STDC__
-#define __weak_reference(sym,alias)     \
+#define __weak_reference_compat(sym,alias)     \
     __asm__(".weak_reference " #alias); \
     __asm__(".set " #alias ", " #sym)
 #else
-#define __weak_reference(sym,alias)     \
+#define __weak_referenc_compate(sym,alias)     \
```
ViralBShah: typo?

gufe44: Thank you.

src/cdefs-compat.h
gufe44: Thank you. Is it okay to check for __weak_reference regardless or should we change this to something like

```
#if defined(__weak_alias) && defined(__NetBSD__)
...
#elif defined(__weak_reference)
...
```

Keno:
Do any of our CI services have NetBSD support? Would be nice to get a CI runner to avoid regressions.

ararslan:
> Do any of our CI services have NetBSD support?

Nope. It _might_ be possible on the paid tier of Cirrus (which allows you to use arbitrary GCP images) but otherwise I don't think so.

gufe44:
Sorry, what I commited is not right at all... is cdefs-compat.h moving in the right direction at least?

gufe44:
Please excuse my "experimental" use of this platform. I really didn't mean to abuse the PR in this way. I am not suggesting https://github.com/JuliaMath/openlibm/compare/master...gufe44:netbsd-fix-openlibm_weak_reference as a solution but it should now somewhat conform to https://github.com/JuliaMath/openlibm/pull/197#pullrequestreview-307404318. Can you verify this is really what you want?

vtjnash:
Looks right to me

ViralBShah:
Make a new PR?

PR #204, opened 2020-03-08
[DO NOT MERGE] log from musl

ViralBShah:
Just for experimenting for anyone who is interested. It appears to be slower than Julia's built-in log, which is in turn slower than the log that ships on mac. This is based on testing from Julia.


ViralBShah:
Also fma is significantly slower if enabled in clang. Yet to try with gcc.

yuyichao:
For fma, did you compiled for correct arch?

ViralBShah:
I assumed the compiler would pick the arch it was building on for starters.

yuyichao:
I think compilers usually don’t do native arch by default

PR #209, opened 2020-06-10
Refactor: OLM_DLLEXPORT definition now lives in a standalone header.

### LINKED ISSUE #191: Create a compiler flag to disable the definition of I
When using this library with C++, there are apparently conflicts with some of the template functions. People are using "I" in C++ headers, which gets substituted with the `_Complex_I` macro. (It's truly a great mystery as to why the particular headers I'm using even include `cmath`, but I'll leave that for another project).

The easiest solution that I can think would be to have a compile-time switch that could disable the definition of `I`. In my initial sanity check, everything seems to compile/link fine - I build openlibm with `I` defined, and then remove the definition for my C++ project's build.

Before I head further down that route (with a potential PR), I wanted to know whether you foresee any difficulties arising from using this library in C++ with `I` undefined (or if you'd accept such a change at all).

phillipjohnston:
On x86_64, openlibm_fenv.h is silently coupled to openlibm_complex.h due to the inclusion of math_private.h inside of openlibm_fenv_amd64.h. The math_private.h header was included to gain access to the OLM_DLLEXPORT macro. However, this macro is redefined, since it is originally defined in openlibm_math.h. This inclusion also exposes math_private.h to public clients, which is undesirable.

Instead, a standalone header is created to contain the OLM_DLLEXPORT definition. This header is used included in the appropriate files. openlibm_fenv_amd64.h is no longer coupled to math_private.h.

Resolves #191

### DIFF
--- a/include/openlibm_defs.h
+++ b/include/openlibm_defs.h
@@ -1,0 +1,14 @@
+#ifndef OPENLIBM_DEFS_H_
+#define OPENLIBM_DEFS_H_
+
+#ifdef _WIN32
+# ifdef IMPORT_EXPORTS
+#  define OLM_DLLEXPORT __declspec(dllimport)
+# else
+#  define OLM_DLLEXPORT __declspec(dllexport)
+# endif
+#else
+#define OLM_DLLEXPORT __attribute__ ((visibility("default")))
+#endif
+
+#endif // OPENLIBM_DEFS_H_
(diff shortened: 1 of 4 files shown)


phillipjohnston:
Hi team, any thoughts on this?

ViralBShah:
I think I missed this notification.

ViralBShah:
Do we want to tag 0.8 as a result of this?

PR #210, opened 2020-07-09
Replace a few remaining __strong_reference uses

Keno:
__strong_alias is an ELF feature that is not available on Darwin/MachO.
We have openlibm_strong_reference to paper over these differences,
but we weren't using it consistently. In particular, we were still
using it to define long double -> double aliases on platforms where
these are the same type (e.g. aarch64). This didn't used to matter,
because the only such platform we supported was always Linux and
thus ELF. This patch fixes these instances thus fixing the build
on Apple Silicon.

### DIFF
--- a/src/cdefs-compat.h
+++ b/src/cdefs-compat.h
@@ -21,7 +21,7 @@
 #define openlibm_strong_reference(sym,aliassym) openlibm_weak_reference(sym,aliassym)
 #else
 #define openlibm_strong_reference(sym,aliassym)	\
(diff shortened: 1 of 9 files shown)


staticfloat:
Looks like 32-bit arm is unhappy.

Keno:
... the order of the symbols is reversed between the two macros. Will fix

Keno:
Looks like the compiler still doesn't like aliasing these with different types? Looks like elswhere we're using weak references which the compiler doesn't seem to be complaining about as much.

PR #214, opened 2020-09-03
Fix Apple Silicon build

Keno:
My previous Apple Silicon build went through, so I thought it already
worked, but it turns out it accidentally built an armv7 build instead.
This actually fixes the Apple Silicon build. One thing to note in
particular is that on Apple Silicong `long double` is the same as
`double` while on Linux `long double` is a 128 bit double-double
format.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -55,6 +55,9 @@ $(error "the mingw32 compiler you are using fails the openblas testsuite. please
 endif
 
 # OS-specific stuff
+ifeq ($(ARCH),arm64)
+override ARCH := aarch64
+endif
 ifeq ($(findstring arm,$(ARCH)),arm)
 override ARCH := arm
 MARCH ?= armv7-a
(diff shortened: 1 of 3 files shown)


Make.inc
```diff
@@ -144,6 +147,13 @@ CFLAGS_add += -I$(OPENLIBM_HOME)/ld128
 endif
 endif
 
+ifneq ($(filter $(ARCH),i387 amd64),)
+LONG_DOUBLE_NOT_DOUBLE := 1
```
ararslan, line 151: Perhaps add a comment about what this means?

Keno: It determines whether `long double` is `double` or not

Keno:
Also cf https://github.com/iains/gcc-darwin-arm64/commit/74b1b621a67db46d91bc9f4885cbabd6b2d240e9

Make.inc
staticfloat:
```suggestion
# Determines whether `long double` is the same as `double` on this arch.
# linux x86_64, for instance, `long double` is 128 bits wide, whereas on macOS aarch64,
# `long double` is the same as `double`. 
LONG_DOUBLE_NOT_DOUBLE := 1
```

staticfloat (approved):

Make.inc
Keno: 80 bits on x86_64 linux, but otherwise sounds good ;)

Keno: unless you're talking about aarch64 for the linux case also

Keno: Ah, I can just edit your comment and then apply it

PR #217, opened 2020-10-23
don't alter toolchain vars if already provided

epsilon-0:
Fixes hard setting of toolchain functions.
Should not affect anyone.

Needed for cross compiling toolchain support.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -10,41 +10,41 @@ VERSION = 0.7.0
 SOMAJOR = 3
 SOMINOR = 0
 DESTDIR =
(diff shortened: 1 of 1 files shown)


ararslan (approved):

epsilon-0:
Any update?

staticfloat:
This should be fine, but note that you can set these values on the command line (e.g. `make prefix=/opt/openlibm`).  The only difference here is that these values will now be inherited from the environment, if they exist.

epsilon-0:
Thanks!

PR #218, opened 2020-11-14
Fix compilation errors on wasm32

jcestibariz:
Fixes missing declaration of `isinff` and invalid library creation due to the use of `ar` instead of `llvm-ar`.


ViralBShah:
These look good enough. merging.

PR #233, opened 2021-05-20
Add missing aliases for 64-bit long double functions

lephe:
Some standard `long double` functions declared in `openlibm_math.h` do not have an implementation in `src`, so they are essentially missing on platforms where `long double` is a `double`.

Specifically, `acoshl`, `asinhl`, `atanhl`, `coshl`, `erfcl`, `erfl`, `expl`, `expm1l`, `log10l`, `log1pl`, `log2l`, `logl`, `powl`, `sinhl`, `tanhl`, and `tgammal`.

This PR uses the `openlibm_weak_reference()` mechanism to define them in terms of the `double`-based version. I'm not certain this is the best approach, but it seems worth a try. I have checked in a cross-compiling environment that all the target functions are exposed and functional.

### DIFF
--- a/bsdsrc/b_tgamma.c
+++ b/bsdsrc/b_tgamma.c
@@ -38,6 +38,7 @@
  * acknowledged.
  */
 
+#include <float.h>
 #include <openlibm_math.h>
 
 #include "mathimpl.h"
@@ -312,3 +313,7 @@ neg_gam(x)
 	if (sgn < 0) y = -y;
 	return (M_PI / (y*z));
 }
+
+#if (LDBL_MANT_DIG == 53)
+openlibm_weak_reference(tgamma, tgammal);
+#endif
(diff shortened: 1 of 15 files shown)


ViralBShah:
It may be worth looking at freebsd's `msun` and seeing what they do. It would be nice to have someone who knows more about the libm spec and innards to suggest if this is a good thing to merge.

lephe:
I'm definitely not qualified to know more about the libm spec (more input welcome), but I did look up the msun source. It turns out (you couldn't make it up!) they do it in exactly the same way. Here is [the alias for `acoshl()`](https://github.com/freebsd/freebsd-src/blob/main/lib/msun/src/e_acosh.c#L66) for example. The two codebases are very similar in that regard, with all long double-related files in msun explicitly supporting only 64-bit and 80-bit mantissas like OpenLibm and leaving the degenerate case to double-related ones.

PR #238, opened 2021-09-07
prefix symbols with _ for 32-bit x86 Windows

### LINKED ISSUE #237: clang i686-w64-windows-gnu weak reference mangling
Sorry, this is probably pretty esoteric...

I am attempting to build for i686-w64-windows-gnu, and it is failing to link.  After some investigation I found the cause.  Clang is hitting this case:
https://github.com/JuliaMath/openlibm/blob/f052f42bb393918acd6fd51ec1f018c5679dfe30/src/cdefs-compat.h#L61-L66

But, on x86 32-bit Windows, symbols for C functions are prefixed by `_`.  Changing that so that `sym` and `alias` are prefixed by `_` solves the linking there.  That doesn't seems like it would work everywhere though.

In contrast, I think GCC would hit the case:
https://github.com/JuliaMath/openlibm/blob/f052f42bb393918acd6fd51ec1f018c5679dfe30/src/cdefs-compat.h#L73-L75

Which always includes `_` prefixes,,,

See also https://github.com/mingw-w64/mingw-w64/blob/6b62cf66832bd158c67662cd5566413baf35f20e/mingw-w64-headers/crt/_mingw_mac.h#L130 and the lines immediately before that figure out whether the `_` is required or not

jeremyd2019:
In a case that I believe can only be hit for Clang i686-*-windows-gnu (AKA MinGW), symbols in asm need to be prefixed with `_`.  Fixes #237

### DIFF
--- a/src/cdefs-compat.h
+++ b/src/cdefs-compat.h
@@ -73,7 +73,7 @@
     __asm__(".weak_reference openlibm_symbol_prefix/**/alias");\
     __asm__(".set openlibm_symbol_prefix/**/alias, openlibm_symbol_prefix/**/sym")
 #endif
-#undef oepnlibm_symbol_prefix
+#undef openlibm_symbol_prefix
 #else	/* !__ELF__ */
 #ifdef __STDC__
 #define openlibm_weak_reference(sym,alias)	\
--- a/src/cdefs-compat.h
+++ b/src/cdefs-compat.h
@@ -59,15 +59,21 @@
 #endif	/* __warn_references */
 #endif	/* __STDC__ */
 #elif defined(__clang__) /* CLANG */
+#if defined(_WIN32) && defined (_X86_)
+#define openlibm_symbol_prefix "_"
+#else
+#define openlibm_symbol_prefix ""
+#endif
 #ifdef __STDC__
 #define openlibm_weak_reference(sym,alias)     \
-    __asm__(".weak_reference " #alias); \
-    __asm__(".set " #alias ", " #sym)
+    __asm__(".weak_reference " openlibm_symbol_prefix #alias); \
+    __asm__(".set " openlibm_symbol_prefix #alias ", " openlibm_symbol_prefix #sym)
 #else
 #define openlibm_weak_reference(sym,alias)     \
-    __asm__(".weak_reference alias");\
-    __asm__(".set alias, sym")
+    __asm__(".weak_reference openlibm_symbol_prefix/**/alias");\
+    __asm__(".set openlibm_symbol_prefix/**/alias, openlibm_symbol_prefix/**/sym")
 #endif
+#undef oepnlibm_symbol_prefix
 #else	/* !__ELF__ */
 #ifdef __STDC__
 #define openlibm_weak_reference(sym,alias)	\


jeremyd2019:
I don't know about the non-`__STDC__` case, or if it's even reachable for Clang i686-w64-windows-gnu

ViralBShah:
@staticfloat Would you be able to take a look at this?

Also, it would be so much easier to merge these if we had CI working through gh actions.

src/cdefs-compat.h
```diff
@@ -59,15 +59,21 @@
...
+    __asm__(".weak_reference " openlibm_symbol_prefix #alias); \
+    __asm__(".set " openlibm_symbol_prefix #alias ", " openlibm_symbol_prefix #sym)
 #else
 #define openlibm_weak_reference(sym,alias)     \
-    __asm__(".weak_reference alias");\
-    __asm__(".set alias, sym")
+    __asm__(".weak_reference openlibm_symbol_prefix/**/alias");\
+    __asm__(".set openlibm_symbol_prefix/**/alias, openlibm_symbol_prefix/**/sym")
 #endif
+#undef oepnlibm_symbol_prefix
```
staticfloat, line 76:
```suggestion
#undef openlibm_symbol_prefix
```

staticfloat:
I talked it over with @vtjnash and we both agree that supporting non-`__STDC__` stuff is a waste of time; very unlikely that anyone is trying to build openlibm with such an archaic C compiler.  Feel free to ignore that, and we will remove the branches in a future commit.

ViralBShah:
This one didn't build on my mac - so I had to revert it.

```
src/s_isinf.c:66:1: error: expected ')'
openlibm_weak_reference(__isinff, isinff);
^
/Users/viral/repos/openlibm/src/cdefs-compat.h:69:32: note: expanded from macro 'openlibm_weak_reference'
    __asm__(".weak_reference " openlibm_symbol_prefix #alias); \
                               ^
src/s_isinf.c:66:1: note: to match this '('
/Users/viral/repos/openlibm/src/cdefs-compat.h:69:12: note: expanded from macro 'openlibm_weak_reference'
    __asm__(".weak_reference " openlibm_symbol_prefix #alias); \
           ^
src/s_isinf.c:66:1: error: expected ')'
openlibm_weak_reference(__isinff, isinff);
^
/Users/viral/repos/openlibm/src/cdefs-compat.h:70:21: note: expanded from macro 'openlibm_weak_reference'
    __asm__(".set " openlibm_symbol_prefix #alias ", " openlibm_symbol_prefix #sym)
                    ^
src/s_isinf.c:66:1: note: to match this '('
/Users/viral/repos/openlibm/src/cdefs-compat.h:70:12: note: expanded from macro 'openlibm_weak_reference'
    __asm__(".set " openlibm_symbol_prefix #alias ", " openlibm_symbol_prefix #sym)
           ^
2 errors generated.
make: *** [src/s_isinf.c.o] Error 1
```

jeremyd2019:
Yeah, it appears that `#undef`ing the symbol prefix does not work (I was worried about that, and my test only succeeded because of the typo).

ViralBShah:
May need a new PR - and while I now  have CI running, I don't think I am able to set up the windows CI correctly for x86.

PR #240, opened 2021-09-08
Update ci.yml for Windows and msys2

ViralBShah:


ViralBShah:
Not sure if I quite have this right, but at least it has CI on all the OSes now.

PR #241, opened 2021-09-09
CI: enable Windows x86

DilumAluthge:
Let's see what breaks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,6 @@ jobs:
             arch: aarch64
           - os: windows-latest
             arch: armv7
-          - os: windows-latest
-            arch: x86
           - os: macos-latest
             arch: armv7
           - os: macos-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,12 @@
 name: OpenLibm CI
-
 on:
   push:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-
+defaults:
+  run:
+    shell: bash
 jobs:
   build:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


DilumAluthge:
cc: @ViralBShah

DilumAluthge:
@ViralBShah You'll need to approve workflows on this PR, since I've never contributed to this repo before.

If you want, you can relax that restriction by going to https://github.com/organizations/JuliaLang/settings/actions and, under the "Fork pull request workflows from outside collaborators" section, click the "Require approval for first-time contributors who are new to GitHub" option and click save.

ViralBShah:
It's hard to tell what is actually breaking. Is it truly getting a 32-bit windows box?

DilumAluthge:
I don't think GHA actually has 32-bit Windows VMs. We'll still be in a 64-bit Windows VM, but we can use a 32-bit compiler.

I'm going to try some of the advice from this thread: https://github.com/actions/virtual-environments/issues/2549

You'll need to re-approve the workflows.

jeremyd2019:
I would recommend using the [setup-msys2 action](https://github.com/msys2/setup-msys2), but I don't know if that's reasonably possible in a single job matrixed with other OSes, I've generally seen a separate 'windows' job with a matrix for the different compilers (32/64-bit gcc/clang).  Also, if you want to be able to build for aarch64 using mingw, https://github.com/mstorsjo/llvm-mingw is probably the best choice.  I've been working on an arm64 environment for msys2, but currently that's all native, which currently means you'd need a self-hosted windows arm64 runner for github actions  ☹️(https://github.com/actions/runner/pull/785 )

DilumAluthge:
I'm hoping that the latest fix I pushed will work. If not, I'll try the setup-msys action. We can choose to run it only on the Windows x86 job.

jeremyd2019:
I'm not sure that relying on mingw toolchains from the pre-installed msys2 is going to continue to work once they update to the 2022 image.  https://github.com/actions/virtual-environments/issues/3949 (bottom of the table)

DilumAluthge:
Ooof. In the short-term, we can just pin to the 2019 image. But you're right, it looks like we'll eventually need to use https://github.com/msys2/setup-msys2 for the 2022 image.

DilumAluthge:
Thanks for the link to that table!

ViralBShah:
Even with setup-msys2, I could only get 64-bit windows working. I'd you see my PR that I merged yesterday it has experimentation and logs with trying to enable msys2.

jeremyd2019:
I think what's happening here is you can't really link with -nodefaultlibs on mingw-w64 (well you *can*, but there's a set of default libs you *need*).  See https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-openlibm/01-mingw-build.patch for what they did to get it building there.

ViralBShah:
Seems like we should adopt that patch from mingw.

jeremyd2019:
I opened #243, don't know if that's preferred.  There are a few test failures that need figuring out though.

PR #242, opened 2021-09-09
prefix symbols with _ for 32-bit x86 Windows

### LINKED ISSUE #237: clang i686-w64-windows-gnu weak reference mangling

jeremyd2019:
Redo of #238
In a case that I believe can only be hit for Clang i686-*-windows-gnu (AKA MinGW), symbols in asm need to be prefixed with `_`.  Fixes #237

### DIFF
--- a/src/cdefs-compat.h
+++ b/src/cdefs-compat.h
@@ -59,14 +59,19 @@
 #endif	/* __warn_references */
 #endif	/* __STDC__ */
 #elif defined(__clang__) /* CLANG */
+#if defined(_WIN32) && defined(_X86_)
+#define openlibm_asm_symbol_prefix "_"
+#else
+#define openlibm_asm_symbol_prefix ""
+#endif
 #ifdef __STDC__
 #define openlibm_weak_reference(sym,alias)     \
-    __asm__(".weak_reference " #alias); \
-    __asm__(".set " #alias ", " #sym)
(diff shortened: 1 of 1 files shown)


ViralBShah:
Can you rename this on master now so that we can get the full CI run?

ViralBShah:
So nice to have the CI. Thanks @jeremyd2019.

PR #243, opened 2021-09-09
msys2 ci

jeremyd2019:
Adds a separate job using setup-msys2 to test all currently available MinGW-w64 environments there.

These currently fail a few tests relating to signgam from lgamma.  I'm not sure if that's due to it calling mingw-w64's variant of lgamma or what.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ name: OpenLibm CI
 
 on:
   push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       - uses: msys2/setup-msys2@v2
         with:
           msystem: ${{matrix.sys}}
-          install: mingw-w64-${{matrix.env}}-toolchain
+          install: base-devel mingw-w64-${{matrix.env}}-toolchain
       - name: Build and run tests
         shell: msys2 {0}
         run: make && make test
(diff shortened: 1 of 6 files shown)


jeremyd2019:
lgamma is in libmingwex, one of the default libs.  But, it should only be pulled in if the symbol is needed by something, and that shouldn't be if openlibm is providing it.

jeremyd2019:
It could also be that it's pulling in mingw's signgam and openlibm's lgamma.  I'll have to investigate it locally to find out exactly what's going on

jeremyd2019:
> It could also be that it's pulling in mingw's signgam and openlibm's lgamma. I'll have to investigate it locally to find out exactly what's going on

That seems to be the case, lgamma is exported from openlibm, but signgam is not.  Will probably have a patch quickly

jeremyd2019:
couple more missing OLM_DLLEXPORTs it looks like

jeremyd2019:
Finally all green!  Sorry for the bit of a mess on this PR.  I can squash some commits, and/or split the CI and fixes into separate PRs if preferred.

jeremyd2019:
oh, the i686 clang fix I made in #242 uses is for `clang32`/`clang-i686`, which is "extra" experimental in that it's not even in the config files by default.  Adding the ability to test that in this CI would require extra special cases just for it.  I don't know if that's something you would care to have in your CI, given that it's so experimental upstream.

ViralBShah:
Don't worry about the squashing. We can squash and merge.

ViralBShah:
I think we don't need to have the experimental support in #242 in CI - just need to make sure it doesn't break anything else.

DilumAluthge:
Thank you @jeremyd2019!

PR #244, opened 2021-09-09
CI: A variety of miscellaneous tweaks

DilumAluthge:


ViralBShah (approved):

.github/workflows/ci.yml
```diff
@@ -23,14 +25,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: Build and run tests
-        run: make && make test
-
-  build-msys2:
+      - run: make
+      - run: make test
+  windows:
     runs-on: windows-latest
```
jeremyd2019, line 34: If you want you can add
```suggestion
    runs-on: windows-latest
    defaults:
      run:
        shell: msys2 {0}
```
And avoid duplicating it between the two steps below

.github/workflows/ci.yml
```diff
@@ -1,12 +1,14 @@
...
 on:
-  push:
   pull_request:
-    branches: [ master ]
-
+    branches:
+      - master
+  push:
+    branches:
+      - master
```
jeremyd2019, line 8: sorry for taking that out if you wanted it.  It made my testing harder on my fork because I wasn't using the 'master' branch there

.github/workflows/ci.yml
DilumAluthge: I thought about that. But for the first step, which sets up the MSYS2 environment, I figured we actually don't want to run that step inside the `msys2` shell. So I figured we didn't want it as a job-wide default.

.github/workflows/ci.yml
DilumAluthge: No worries!

.github/workflows/ci.yml
jeremyd2019: We do that over at msys2, and it works fine.  That default shell is only used by `run` steps, not actions (which are actually run in node,js)

DilumAluthge: Oh interesting, I had always assumed that the default shell would also be used by actions.

I wonder what happens in composite actions (i.e. non-Javascript actions).

DilumAluthge: Let's give it a try - https://github.com/JuliaMath/openlibm/pull/247

jeremyd2019: I don't know.  Composite actions are newer than the msys2 CIs I've dealt with

jeremyd2019: This is getting pretty far afield, but https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsshell says shell is *required* for run steps in a composite action, so I guess the default shell doesn't apply.

Issue #245, opened 2021-09-09
CI architectures


DilumAluthge:
So, adding the `arch` variable to your build matrix doesn't actually give you a VM with that architecture. The only thing that determines what kind of VM you get is the `runs-on: ${{ matrix.os }}` key.

Details on the available VMs are here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

So e.g. `ubuntu-latest` gets you an Ubuntu x86_64 VM. `macos-latest` gets you a macOS x86_64 VM. And `windows-latest` gets you a Windows x86_64 VM.

If you want to e.g. build an x86 build and test it CI, I think you need to install the appropriate 32-bit compiler, and then pass the `ARCH` make variable. 

cc: @ViralBShah @jeremyd2019

DilumAluthge:
For Windows, thanks to @jeremyd2019's work, I think we are doing this correctly, because we are correctly setting up the desired MSYS2 environment. So I think on Windows, we are indeed doing 32-bit building and testing when requested.

But currently, for Linux and macOS, we are always doing x86_64, regardless of the value of the `arch` variable.

ViralBShah:
Thanks. I think I don't want to do any more work on CI - I am satisfied with what we have. If someone wants to enable linux 32-bit CI, that will be nice to have. I feel we have adequate coverage.

How do we get arm/aarch64 support going though? I see https://github.com/marketplace/actions/run-on-archlinuxarm. I used to think that github actions natively supported arm.

jeremyd2019:
Yes, on Windows you are getting proper 32 and 64-bit x86 builds through the `msystem` setting on the setup-msys2 action.  You could theoretically do a cross-compile for Windows aarch64 (or ARM64 as they call it) using llvm-mingw, but you couldn't test it without arm hardware.

jeremyd2019:
> How do we get arm/aarch64 support going though? I see https://github.com/marketplace/actions/run-on-archlinuxarm. I used to think that github actions natively supported arm.

Nope ☹️   All they've got is the ability to run a self-hosted runner on linux aarch64.  No hosted-by-them runners, and no Windows aarch64 (https://github.com/actions/runner/pull/785 has been open quite a while)

DilumAluthge:
I agree that for now, I think we are fine with what we have.

jeremyd2019:
[...]

PR #246, opened 2021-09-09
CI (Linux and macOS): Remove the `arch` variable, which currently has no effect

DilumAluthge:
Ref #245

The presence of the `arch` variable is misleading, because it currently has no effect.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,16 +16,6 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
-        arch:
-          - x64
-          - x86
-          - armv7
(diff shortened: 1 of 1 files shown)


ViralBShah:
Yeah, I was wondering if github actions somehow knew something about `arch`. How do we test on Linux 32-bit, for example?

ViralBShah:
Why is it running everything twice now? push and pull_request?

ViralBShah (approved):

DilumAluthge:
> Why is it running everything twice now? push and pull_request?

Yeah. I've fixed that in https://github.com/JuliaMath/openlibm/pull/244.

PR #248, opened 2021-09-09
Bump version to 0.8

ViralBShah:
Bump version to 0.8 since we now have working CI now, with proper platform support, arm64 on win, and assorted fixes.

Bump the SOVERSION as well since we have some new exports and such.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -6,8 +6,8 @@ default: all
 OS := $(shell uname)
 # Do not forget to bump SOMINOR when changing VERSION,
 # and SOMAJOR when breaking ABI in a backward-incompatible way
-VERSION = 0.7.0
-SOMAJOR = 3
+VERSION = 0.8.0
+SOMAJOR = 4
 SOMINOR = 0
 DESTDIR =
 prefix ?= /usr/local


kargl:
If you're going to bump the version number, can I suggest that you look at issues 205, 206, and 206.  Fairly, trivial patches.

ViralBShah:
Any chance you have patch files for these handy? 

I guess we can also keep the 0.7 series for now.

kargl:
I barely know how to use git.  I can clone things from github. I probably had patches 1.5 years ago when I opened the issues.  Unfortunately, no one ever acknowledged the issues.  So, things rot and get lost.  I likely found the issues by comparisons between FreeBSD's libm and Openlibm.  Openlibm is derived from FreeBSD.  A diff might be fruitful.

ViralBShah:
Sorry about the silence. I periodically collect your patches and apply them.

In general in Julia, we no longer use openlibm for much anymore. I just like to keep this project going since it is valuable thave it as an independent libm.

ViralBShah:
Nite: need to rebase to master for win64 arm.

ViralBShah:
I'm going to merge this today. We can always go to newer versions - 0.9, 0.10, 0.11, etc. as other fixes pour in.

nalimilan:
New exports don't require a new SOMAJOR bump.

ViralBShah:
Do they require an SOMINOR bump?

nalimilan:
Yes. But bumping SOMINOR isn't a problem. What's annoying is that bumping SOMAJOR forces distributions to recompile all packages that depend on openlibm.

ViralBShah:
I can never remember what to do when. Would be nice to clearly note in the Makefile.

ViralBShah:
Should we make another release resetting the SOMAJOR and bumping the SOMINOR?

nalimilan:
> I can never remember what to do when. Would be nice to clearly note in the Makefile.

Well it already says:
```
# Do not forget to bump SOMINOR when changing VERSION,
# and SOMAJOR when breaking ABI in a backward-incompatible way
```
This is the same as semantic versioning actually.

> Should we make another release resetting the SOMAJOR and bumping the SOMINOR?

What we could do is make more releases with the old SOMAJOR version (3), and the day we break the API, skip SOMAJOR 4 (which has already been used in the release following this PR) and go directly to 5. But I'm not sure it would be worth it, and we're likely to forget to skip 4, which would really be problematic.

nalimilan:
Yeah I guess there's no better choice... :-/

PR #251, opened 2021-09-10
Fixes for Windows ARM64

### LINKED ISSUE #250: does not build on Windows ARM64
https://github.com/JuliaMath/openlibm/blob/f88e52a89eb162d2fc0a8abd3ec2c9b6aeea16c3/src/aarch64_fpmath.h#L30-L44

Results in errors because `unsigned long` is 32-bits on Windows.  Could this use `uint64_t`, or does this need some sort of `__LP64__` check like https://github.com/JuliaMath/openlibm/blob/f88e52a89eb162d2fc0a8abd3ec2c9b6aeea16c3/src/fpmath.h#L34-L39 (though I don't see a 32-bit arm equivalent like the x86 case).

After hacking around that I get
```
src/s_cosl.c:44:2: error: "Unsupported long double format"
```

jeremyd2019:
With these changes, it builds and tests succeed on Windows ARM64.  Fixes #250

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -151,7 +151,7 @@ ifneq ($(filter $(ARCH),i387 amd64),)
 # `long double` is the same as `double`. 
 LONG_DOUBLE_NOT_DOUBLE := 1
 else ifeq ($(ARCH), aarch64)
-ifneq ($(OS),Darwin)
+ifeq ($(filter $(OS),Darwin WINNT),)
 LONG_DOUBLE_NOT_DOUBLE := 1
 endif
 endif
--- a/src/aarch64_fpmath.h
+++ b/src/aarch64_fpmath.h
@@ -30,15 +30,15 @@
 union IEEEl2bits {
 	long double	e;
 	struct {
-		unsigned long	manl	:64;
-		unsigned long	manh	:48;
+		uint64_t	manl	:64;
+		uint64_t	manh	:48;
 		unsigned int	exp	:15;
 		unsigned int	sign	:1;
 	} bits;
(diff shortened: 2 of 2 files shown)


jeremyd2019:
Oops, this worked locally, but when applied to MINGW-packages it fails with:
```
  In file included from src/s_fmax.c:32:
  In file included from C:/_/M/mingw-w64-openlibm/src/build-aarch64/src/fpmath.h:33:
  C:/_/M/mingw-w64-openlibm/src/build-aarch64/src/aarch64_fpmath.h:33:3: error: unknown type name 'uint64_t'
                  uint64_t        manl    :64;
                  ^
  C:/_/M/mingw-w64-openlibm/src/build-aarch64/src/aarch64_fpmath.h:34:3: error: unknown type name 'uint64_t'
                  uint64_t        manh    :48;
                  ^
  C:/_/M/mingw-w64-openlibm/src/build-aarch64/src/aarch64_fpmath.h:40:3: error: unknown type name 'uint64_t'
                  uint64_t        manl    :64;
                  ^
  C:/_/M/mingw-w64-openlibm/src/build-aarch64/src/aarch64_fpmath.h:41:3: error: unknown type name 'uint64_t'
                  uint64_t        manh    :48;
                  ^
  4 errors generated.
  make: *** [Make.inc:160: src/s_fmax.c.o] Error 1
```

Maybe there are more changes in master from 0.7.5 that allowed it to work?

jeremyd2019:
Happens on master too.  I swear I tested this, but maybe only incrementally as I was changing things.

PR #253, opened 2021-09-10
Another Windows ARM64 fix

jeremyd2019:
include stdint.h to be sure that uint64_t can be used.
This header already had references to uint32_t, so I assumed it was safe to use those typedefs, but a compile error revealed it was not always.

### DIFF
--- a/src/aarch64_fpmath.h
+++ b/src/aarch64_fpmath.h
@@ -25,7 +25,7 @@
  * SUCH DAMAGE.
  *
  * $FreeBSD: head/lib/libc/aarch64/_fpmath.h 281197 2015-04-07 09:52:14Z andrew $
-*/
+ */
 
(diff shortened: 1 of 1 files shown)


src/aarch64_fpmath.h
```diff
@@ -25,7 +25,9 @@
  * SUCH DAMAGE.
  *
  * $FreeBSD: head/lib/libc/aarch64/_fpmath.h 281197 2015-04-07 09:52:14Z andrew $
- */
+*/
```
jeremyd2019, line 28:
```suggestion
 */
```

ViralBShah:
Revert this then?

jeremyd2019:
Yeah it *actually* builds from a clean build.  I don't know why it didn't fail when I was working on the prior PR.

PR #255, opened 2021-11-10
Add fmod assembly version

### LINKED ISSUE #93: fmod for 64bit is slow
The `fmod` function on `x86_64` is pretty slow (e.g. roughly 5x slower than `remainder`, or `fmod` in the os x libm). From what I understand of the Makefile, it is using the `src/e_fmod.c`, which is implemented in software. The other remainder functions (`remainder`, `remquo`) use the assembly implementations in `amd64` (which use x87 instructions).

I should mention that this isn't a problem for Julia at the moment, as we access `fmod` through the LLVM intrinsic, which in turn calls the system libm. But if this were to change, we would see a significant performance hit.

ViralBShah:
From https://github.com/freebsd/freebsd-src/tree/master/lib/msun/amd64

Fix #93 

cc @simonbyrne


ViralBShah:
@yuyichao Does this look right?

PR #286, opened 2024-01-08
Add CMake support

ryonaldteofilo:
I've been using this library to achieve consistent floating-point maths in cross-platform multiplayer projects at work. We use CMake to cross-compile across different platforms. Thought it would be useful to add this so other can easily integrate the library to their projects as well :)


ViralBShah:
While we will continue with the existing build system, happy to include a cmake build as well. Can you make a PR to the README mentioning it and how to generate shared libraries with the cmake build?

ryonaldteofilo:
ah yea I've just opened a pr adding CMake build instructions on the README and additional changes to allow building it as a shared lib

PR #293, opened 2024-01-12
Correct long double check in CMakeLists.txt and remove duplicate definitions for static library builds too

### LINKED ISSUE #292: CMake build fails on macOS arm64/aarch64
Issue seems to be caused by a mistake in the `CMakeLists.txt` file, specifically the `long double == double` check.

ryonaldteofilo:
- Fixed cmake aarch64 build for macOS due to a mistake in `long double == double` size check, this should fix #292 .
- Apply the filtering out of C implementation if native implementation exists for static library builds as well.
- Default to building a shared library just like the GNU make build, and updated the README to reflect this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ project(openlibm
         VERSION 0.8.0
         LANGUAGES C ASM)
 
-add_library("${PROJECT_NAME}" STATIC)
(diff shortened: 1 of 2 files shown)


ryonaldteofilo:
Probably should mention that I've confirmed that this indeed does fix #292 :)

PR #298, opened 2024-02-02
test: add musl-libc testsuite

inkydragon:
Do NOT merge with squash commits!

- `make coverage` will run glibc and musl test, and then gen report.
	- Currently, it passes all musl tests under linux CI. 
	    Other platforms have not been tested yet. Wait for the following pr
- `make test` will only run glibc test.
	- TODO: run musl test on win and mac CI
- All modifications are marked with `XXX`.
- Tests that are currently skipped are marked with `TODO`.
	- Some tests were moved to subfolders to skip them
	    `src\math\broken\*`, `src\math\not-impl\*`

Since there are a lot of new files added, it is recommended to review the changes made after the addition of new tests one by one.

-----

broken tests:
- fenv
- fma
- fmaf
- fmal
- fpclassify
- j1
- j1f
- jn
- jnf
- nearbyintl
- nextafterl
- nexttoward
- nexttowardl
- round
- tgamma
- yn
- ynf

not impl funcs:
- drem
- dremf
- exp10
- exp10f
- exp10l
- pow10
- pow10f
- pow10l
- scalb
- scalbf


ViralBShah:
Just thinking out aloud, if the musl libm passes all these tests - would it be best to pull out the libm from musl into a standalone library?

inkydragon:
> pull out the libm from musl into a standalone library 
> and update openlibm to that?

I like this idea.
Moving a step further, we could take the libm related tests from libc-testsuite (for musl)
and put them in a standalone project, maybe call it `openlibm-test`.
Instead of adding all these things to `openlibm`.

Then, we have:
- openlibm == musl fork
- openlibm-test == libc-testsuite fork

----

@ViralBShah 
Another thing I am thinking about is after julia removes its dependency on openlibm (and then links to the system libm by default), would it be more useful to have a libm written by julia than to continue to improve the precision of openlibm?

ViralBShah:
It's a good idea to have a separate repo for the tests. Julia almost has no dependency on openlibm, except in a few cases where LLVM on win32 needs it for a couple of things.

If the tests are in a separate repo, we can recreate a libm from Julia and then use the testsuite to test the Julia implementation.

PR #299, opened 2024-03-15
CMake Changes

theoparis:
- Removing SHARED allows openlibm to be built statically depending on the value of BUILD_SHARED_LIBS
- Using CMAKE_CURRENT_SOURCE_DIR allows openlibm to be built via FetchContent
- I added the riscv64 architecture to openlibm's cmake build scripts


ViralBShah:
Merge?

ViralBShah:
Can you rebase to master? Is this ok to merge given the changes in #302? It looks like there is no conflict.

ViliusSutkus89:
Hello @theoparis , could you confirm that the current git version works for you with FetchContent?

Also, if you don't mind me asking, have you tried using package managers instead of FetchContent? I'm assuming you're using it to pull OpenLibm into your project

PR #303, opened 2024-06-19
Minor fix-ups for Darwin PowerPC

barracuda156:
@ViralBShah This took forever :) Sorry for a ridiculous delay, it was forgotten.

Here are two tiny fixes:

1. Use a macro which includes `ppc64`, so that it is not left broken on macOS. (BTW, this still leaves out AIX, I believe?)
2. Make sure `__BYTE_ORDER__` is actually defined: AFAIK, it may not be the case with some old gcc versions, and those support C99, so it is relevant.

### DIFF
--- a/include/openlibm_fenv.h
+++ b/include/openlibm_fenv.h
@@ -8,7 +8,7 @@
 #include <openlibm_fenv_amd64.h>
 #elif defined(__i386__)
 #include <openlibm_fenv_i387.h>
-#elif defined(__powerpc__) || defined(__ppc__)
+#elif defined(__powerpc__) || defined(__POWERPC__)
 #include <openlibm_fenv_powerpc.h>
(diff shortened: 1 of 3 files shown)


ViralBShah:
Windows is trying to use gcc instead of clang. Otherwise, happy to merge.

barracuda156:
@ViralBShah Thanks for merging! This was super-fast.

ViralBShah:
Would you like me to tag a new release?

barracuda156:
> Would you like me to tag a new release?

That would be certainly helpful, thank you.

ViralBShah:
Released 0.8.3

barracuda156:
> Released 0.8.3

Updated in MacPorts in https://github.com/macports/macports-ports/commit/9e4787078b3f00708932da4214d9642c3e458302

Issue #301, opened 2024-06-17
CMake parity with Makefile


Zopolis4:
The CMake build option does not seem to have support for installing the built library, nor does it produce both static and shared libraries. Additionally, it also does not create an `openlibm.pc` file, nor does it seem to have support for installing headers.

ViliusSutkus89:
Hello @Zopolis4 , everything you've mentioned except `openlibm.pc` would be handled by #302

Zopolis4:
Thanks! It's certainly a step in the right direction, but I do need the `openlibm.pc` file to be able to switch buildsystems. I can't re-open the issue myself, though.

PR #304, opened 2024-06-21
CMake changes: Long double/double size detection, Android x86 and armv7 support, install pkg-config pc file

ViliusSutkus89:
Hello,

This PR has the following changes:
- Generate pkg-config openlibm.pc file (issue #301)
- Support for Android armv7-a and Android x86
- Reworked long double size detection. Previously it was hardcoded, I've added a test which compares LDBL_MANT_DIG and DBL_MANT_DIG from float.h. Previously hardcoded cases did not cover Android x86 correctly. A simple compiler check should produce a reliable result.

Also, when we release, we need to update version number at the top of CMakeLists.txt file, otherwise the generated pkg-config openlibm.pc file will report the wrong version. I've set the version to 0.8.4 in advance. I don't really care about it, because I don't use pkg-config.pc file, I'm just saying what would be the consequences of not incrementing the version number.
Where do we set the version number for pkg-config.pc file generated by autotools?

Once again, #299 will require a merge commit, because it's the same file we are editing, but we touched different lines, so we should be good.

Regards


ViralBShah:
@ViliusSutkus89 Happy to give you commit access here. Please let me know if ok.

ViliusSutkus89:
Sure. I can't be a full time maintainer, but I'll try to help whenever I can

ViliusSutkus89:
Updated PR with a patch to match version in CMake with the version in Autotools

Issue #278, opened 2023-06-01
aarch64/fenv.c:41:23: error: invalid initializer


ghost:
Trying to build this for aarch64, and even tried to remove the constant to fix this, but I think I'll open an issue on this. I hope this is just me compiling this incorrectly haha

syscall7788:
![Screenshot from 2023-05-31 20-16-45](https://github.com/JuliaMath/openlibm/assets/134033388/202ed707-2595-43ea-a7e0-fe215c7d746a)

Apparently even on older releases this still occurs.

000lbh:
since fenv_t is a struct, its initializer should be {0}. You can change this line into:
const fenv_t __fe_dfl_env = {0};
However, I have no idea why the original code compiles. On my machine of loongarch64 architecture, it doesn't either.

giordano:
There's the same warning on riscv64:
```
In file included from /home/runner/work/openlibm/openlibm/include/openlibm_fenv.h:18,
                 from src/s_lrint.c:29:
In function ‘feupdateenv’,
    inlined from ‘lrint’ at src/s_lrint.c:59:2:
/home/runner/work/openlibm/openlibm/include/openlibm_fenv_riscv.h:87:25: warning: ‘env’ is used uninitialized [-Wuninitialized]
   87 | #define __wfs(__fcsr)   __asm __volatile("csrw fcsr, %0" :: "r" (__fcsr))
      |                         ^~~~~
/home/runner/work/openlibm/openlibm/include/openlibm_fenv_riscv.h:213:9: note: in expansion of macro ‘__wfs’
  213 |         __wfs(*__envp);
      |         ^~~~~
src/s_lrint.c: In function ‘lrint’:
src/s_lrint.c:52:16: note: ‘env’ was declared here
   52 |         fenv_t env;
      |                ^~~
```
This also makes Julia crash on riscv64 during bootstrap after the [update of openlibm to v0.8.4](https://github.com/JuliaLang/julia/commit/5ec41e90c20ee3825d32d2d374920ce0df81f73c) with the following error message
```
$ make -j3
[...]
    JULIA usr/lib/julia/sys-o.a
Collecting and executing precompile statements
└ Collect (Basic: ✓ fatal: error thrown and no exception handler available.
BoundsError(a=Array{Base.Partr.taskheap, 1}(dims=(0,), mem=Memory{Base.Partr.taskheap}(0, 0x3f848c0a60)[]), i=(0,))
[...]

PR #308, opened 2024-10-10
Fix fenv_t initializer

000lbh:
I have no idea why previous versions compile on GitHub actions, but they don't on my loongarch64 machine, and there was also someone reported on aarch64 architecture #278 . Initializer of a struct should be a brace-enclosed list. Btw, my gcc version is 14.2.1, it seems that cross-compile gcc version in GitHub actions is old.

### DIFF
--- a/loongarch64/fenv.c
+++ b/loongarch64/fenv.c
@@ -1,5 +1,5 @@
 #define	__fenv_static
-#include <fenv.h>
+#include <openlibm_fenv.h>
 
 #ifdef __GNUC_GNU_INLINE__
 #error "This file must be compiled with C99 'inline' semantics"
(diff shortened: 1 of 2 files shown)


yuyichao:
It seems that the definition of `fenv_t` in this repo is **not** a struct except on x86 (both 32 and 64bit). However, the system one does seem to be a `struct`. Which one is intended to be used?

000lbh:
Thanks for you comment. I didn't notice there are some openlibm_fenv_ARCH.h in the repo before. It is strange that fenv.c of some architectures like arm, amd64 and aarch64 includes openlibm_fenv.h, which uses project's fenv.h, but others like loongarch64 and riscv64 directly includes system's fenv.h

000lbh:
Maybe this version is ok

PR #311, opened 2024-12-14
ci: fix clang build

inkydragon:
> The active environment is selected via the `MSYSTEM` environment variable
>
> https://www.msys2.org/docs/environments/


inkydragon:
Replaced by #312

PR #313, opened 2025-01-05
ci: add loongarch64

inkydragon:
- The `gcc-loongarch64-linux-gnu` package seems not to exist on Ubuntu 24.04, so install `gcc-14-loongarch64-linux-gnu` manually.
	See: https://github.com/JuliaMath/openlibm/actions/runs/12617775708/job/35160409682?pr=313#step:3:55
- Manual specification of `CC, AR` during compilation and testing


inkydragon:
Seems we need `ubuntu-24.10` rather than `ubuntu-24.04 (LTS)`

https://pkgs.org/download/gcc-loongarch64-linux-gnu

ViralBShah:
Please merge when you think it is ready

PR #314, opened 2025-01-07
Fixes for building on AArch64 with Clang

ararslan:
See individual commit messages for descriptions and motivations of changes. We should tag a new patch version after this is merged.


giordano (approved):
Looks good to me as far as I can tell, the `ifeq` test was definitely wrong (it could never be true)

PR #316, opened 2025-01-10
Initialize all `fenv_t` with `__fe_dfl_env`

### LINKED ISSUE #278: aarch64/fenv.c:41:23: error: invalid initializer

giordano:
Opening as a draft, but the hope is to fix #278.

### DIFF
--- a/src/e_sqrtl.c
+++ b/src/e_sqrtl.c
@@ -81,7 +81,7 @@ sqrtl(long double x)
 	union IEEEl2bits u;
 	int k, r;
 	long double lo, xn;
-	fenv_t env;
+	fenv_t env = __fe_dfl_env;
 
 	u.e = x;
 
--- a/src/s_lrint.c
+++ b/src/s_lrint.c
@@ -49,7 +49,7 @@
 OLM_DLLEXPORT dtype
 fn(type x)
 {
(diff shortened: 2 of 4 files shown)


giordano:
At least CI here is happy, as much that means 🙂  Now I'm recompiling Julia on https://github.com/JuliaLang/julia/commit/3c7bc45f14c8a2b70ccf1d92d4418cfaaa112b56 with this change, to see if this is enough to fix compilation failure mentioned at https://github.com/JuliaMath/openlibm/issues/278#issuecomment-2581246607.

giordano:
Unfortunately this PR doesn't solve the issue reported at https://github.com/JuliaMath/openlibm/issues/278#issuecomment-2581246607 😞

> Instead of initializing env in the first place, can you try implementing fenv_t loading in feholdexcept(), which is a no-op on RISC-V?

I'm not quite sure what you mean (sorry, I'm not really familiar with openlibm nor riscv-v internals, I'm just trying to fix the issue I observed).  Would you be able to open a PR?  Or at least give me some more directions of what to do?

PR #318, opened 2025-01-10
Add Codecov token to CI

ararslan:
Coverage hasn't been properly submitted in a long time. The issue as reported by the CI job is that we need a token, so I've added one in the repository settings and this commit just tells the action to use it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           files: ./cov-html/libopenlibm.info
+          token: ${{ secrets.CODECOV_TOKEN }}
       - uses: actions/upload-artifact@v4
         with:
           name: code-coverage-report


giordano:
>  I've added one in the repository settings and this commit just tells the action to use it.

For what is worth, there's already an organisation-wide one, you overrode that one.

ararslan:
Was the organization-wide token there in early January? Shouldn't matter regardless whether the organization-wide or repository-level token is used.

giordano:
It [has been there](https://github.com/JuliaMath/openlibm/settings/secrets/actions) for a year :wink:

ararslan:
Huh, I wonder why Codecov was complaining in CI about the token then. Maybe the action version was old or something. This was ~2 months ago so has been pretty well GC'd from my brain.

giordano:
I think you only needed the change in this PR, without creating a new token 🙂

ararslan:
Ahhhh, that makes sense. Feel free to delete the repo token if you prefer.

PR #322, opened 2025-03-11
ci: setup ClangArm64 CI

inkydragon:


inkydragon:
> /bin/sh: line 1: /clangarm64/bin/clang: cannot execute binary file: Exec format error

See: #323

PR #324, opened 2025-05-07
riscv: Fix feholdexcept()

AlekseyZhmulin:
The feholdexcept() function must store the current floating point environment in *__envp.
Related to #321


inkydragon (commented):
We're doing the same thing as BSD.
I'm not sure this works.

https://github.com/freebsd/freebsd-src/blob/master/lib/msun/riscv/fenv.h#L188-L195

AlekseyZhmulin:
The problem is that feholdexcept() does not save floating point environment, but feupdateenv() loads floating point environment. Therefore, in the lrint() function, "fenv_t env" is not initialized and a random value from the stack is loaded into the fcsr register. https://github.com/JuliaMath/openlibm/blob/master/src/s_lrint.c#L59

ViralBShah:
Musl: https://git.musl-libc.org/cgit/musl/tree/src/fenv/feholdexcept.c?id=6f2e5607d2d463dfc342489660e089553b749dcf

PR #344, opened 2026-06-14
enhance CMake script and add CMake to CI tests

scivision:
Use YAML anchors and composite workflow to deduplicate Github Actions scripts

Modernize CMake script and clarify by contemporary syntax


ViralBShah:
Good to merge?

PR #351, opened 2026-06-22
Fix powl() returning NaN instead of +0 on extreme underflow (#334)

### LINKED ISSUE #334: powl yields NaN instead of +0
for x=0x1.98p-3072l and y=0xa.ab43b6dba9a6383p+16364l, powl yields NaN instead of +0 (even with only one thread, cf
https://github.com/JuliaMath/openlibm/issues/222)

zimmerma@coriandre:~/svn/tbd/20/src/binary80$ cat test2_pow_openlibm.c
#include <stdio.h>
#include <math.h>
#include <omp.h>
 
 
int main()
{
  long double x = 0x1.98p-3072l;
  long double y = 0xa.ab43b6dba9a6383p+16364l;
  long double z;
  z = powl (x, y);
  printf ("z=%La\n", z);
  return 0;
}

zimmerma@coriandre:~/svn/tbd/20/src/binary80$ gcc -fno-builtin test2_pow_openlibm.c /localdisk/zimmerma/openlibm-0.8.7/libopenlibm.a
zimmerma@coriandre:~/svn/tbd/20/src/binary80$ ./a.out
z=-nan

zimmerma@coriandre:~/svn/tbd/20/src/binary80$ gcc -fno-builtin test2_pow_openlibm.c -lm # with GNU libc
zimmerma@coriandre:~/svn/tbd/20/src/binary80$ ./a.out
z=0x0p+0

ViralBShah:
Fixes #334.

`powl(0x1.98p-3072L, 0xa.ab43b6dba9a6383p+16364L)` returned `NaN`; glibc returns `+0` (underflow).

## Root cause
In `ld80/e_powl.c`, `powl` forms `G = y*log2(x)` in 1/NXT units (here `G ≈ -2^16379`). The reduction helper `reducl(G)` computes `ldexpl(G, LNXT) = G*32 ≈ -2^16384`, which **overflows to -Inf**. That cascades (`Gb = G - (-Inf) = +Inf`; `w = ldexpl(-Inf + +Inf, …) = NaN`), so the over/underflow tests `w > MEXP` / `w < MNEXP` are both false and the routine falls through returning NaN. The true result underflows to +0. The bug is shared verbatim with current FreeBSD/OpenBSD msun, so there is no upstream patch to import.

## Fix
Short-circuit over/underflow on `G` (which already holds the final exponent to within O(1)) *before* `reducl(G)` can overflow, with a generous `2*MEXP/NXT` margin so genuine borderline values still reach the original precise test:
```c
G = Fa + w * ya;
if (G >  2.0L*MEXP/NXT)  return huge*huge;          /* overflow  */
if (G <  2.0L*MNEXP/NXT) return twom10000*twom10000;/* underflow */
Ga = reducl(G);
```

## Verification
- Reproducer now returns `0x0p+0`, matching glibc.
- 19-case battery (normal, overflow→Inf, underflow→+0, negative bases, non-integer exponents, `1^y`, `x^0`, and four over/underflow **boundary** cases at y=16383.5/16384.5/−16445/−16500) all match glibc, including exact boundary classification.
- Full `make test` passes (double + float + exception-flag suites).
- New `test/regression/test-334.c` asserts the result is `+0` and not NaN; verified it **fails before** this fix and **passes after** (skips where `long double == double`).

Stacked on the regression-harness PR #350 (provides `test/regression/`).

### DIFF
--- a/ld80/e_powl.c
+++ b/ld80/e_powl.c
@@ -404,6 +404,27 @@ Fa = reducl(F);
 Fb = F - Fa;
 
 G = Fa + w * ya;
+
+/*
+ * Guard against intermediate overflow in reducl(G).  When |y*log2(x)|
+ * is enormous, reducl(G) computes ldexpl(G, LNXT) which overflows to
+ * +-Inf; the subsequent Inf-Inf / Inf+(-Inf) arithmetic then yields
+ * NaN, which defeats the w > MEXP / w < MNEXP tests below and makes
+ * powl() return NaN instead of Inf (overflow) or +0 (underflow).
+ * G holds y*log2(x) in 1/NXT units, so the final exponent is
+ * ldexpl(G, LNXT) to within O(1).  Genuine over/underflow occurs at
+ * |G| ~ MEXP/NXT ~ 16448, vastly below the |G| ~ LDBL_MAX/NXT at
+ * which reducl(G) overflows.  A generous margin (2*MEXP/NXT) keeps
+ * this guard well clear of the real boundary, so borderline cases
+ * still fall through to the precise test on w = ldexpl(Ga+Ha, LNXT)
+ * below, while the pathological huge-|G| case is caught here before
+ * reducl(G) can overflow.
+ */
+if( G > (2.0L * MEXP / NXT) )
+	return (huge * huge);		/* overflow */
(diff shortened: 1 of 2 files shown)


zimmermann6:
good work! I'll check that with the next release

PR #355, opened 2026-06-22
Make powl() thread-safe (#222)

### LINKED ISSUE #222: powl returns non-sense results in multi-thread mode
with the attached program I get with OpenLibm 0.7.4:

$ gcc -fno-builtin test_pow_openlibm.c /localdisk/zimmerma/openlibm-0.7.4/libopenlibm.a -fopenmp

$ OMP_NUM_THREADS=1 ./a.out
nthreads=1

$ OMP_NUM_THREADS=2 ./a.out
nthreads=2
Error for powl(0x8.779021e7c2f81b2p+14982,-0x8.0021b03e1f821c9p-15097), expected 1, got 0x8.8980e8092da8445p-3
Error for powl(0x8.779021e7c2f81b2p+14982,-0x8.0021b03e1f821c9p-15097), expected 1, got 0x8.2fae3a785f61333p-3

There seems to be an issue with powl when several threads use it.

[test_pow_openlibm.c.gz](https://github.com/JuliaMath/openlibm/files/5923856/test_pow_openlibm.c.gz)

ViralBShah:
Fixes #222.

`powl` returned garbage under concurrent calls — e.g. two threads computing `powl(0x8.779021e7c2f81b2p+14982L, -0x8.0021b03e1f821c9p-15097L)` (exactly 1) got different huge/Inf values.

## Root cause
`ld80/e_powl.c` kept `powl()`'s working scratch in **file-scope `static` mutable** variables:
```c
static volatile long double z;
static long double w, W, Wa, Wb, ya, yb, u;
```
Concurrent calls race on this shared state. (`reducl()`/`powil()` already used locals; the read-only coefficient tables are fine.)

## Fix
Move that scratch to local automatic variables inside `powl()` (preserving the load-bearing `volatile` on `z` as a local). No math changes.

## Verification
- 8-thread stress (each thread computing a distinct exactly-representable power, including the issue value): **fails every run before**, **rock solid after**.
- Single-threaded parity vs glibc on a 21-value battery (normal/overflow/underflow/negative base/non-integer exponent) — unchanged.
- Full `make test` passes.
- New `test/regression/test-222.c` (pthreads; ld80-only via `#if LDBL_MANT_DIG == 64`, so it builds/runs natively and compile-skips elsewhere). Proven to fail before / pass after.

Note: touches `ld80/e_powl.c` in a different region than #351 (#334's overflow guard); no conflict.

### DIFF
--- a/ld80/e_powl.c
+++ b/ld80/e_powl.c
@@ -189,8 +189,6 @@ static long double R[] = {
 static const long double MAXLOGL = 1.1356523406294143949492E4L;
 static const long double MINLOGL = -1.13994985314888605586758E4L;
 static const long double LOGE2L = 6.9314718055994530941723E-1L;
-static volatile long double z;
-static long double w, W, Wa, Wb, ya, yb, u;
 static const long double huge = 0x1p10000L;
 #if 0 /* XXX Prevent gcc from erroneously constant folding this. */
 static const long double twom10000 = 0x1p-10000L;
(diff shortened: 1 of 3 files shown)


zimmermann6:
good work! I'll check that with the next release

